PR #2, opened 2017-09-22
Add `is.class()`

SchneiderOr:


sindresorhus:
When would you need to know if something is a `class`? I'm genuinely curious. I considered adding this initially, but couldn't think of when I would use it,

sudo-suhas:
Maybe to see if `new` was required for invoking?

SchneiderOr:
In my use case i have a tool that parsing react components to extract their displayName, propTypes, defaultProps etc... based on some strategies, one of them is to check wether the component you intent to extract props from is an ES6 class instance or a functional component and then handle the parsing and extracting of the data differently.

test.js
```diff
@@ -130,6 +134,10 @@ test('is.array', t => {
 	testType(t, 'array');
 });
 
+test('is.class', t => {
+	testType(t, 'function');
```
sindresorhus: `function` => `class`

index.js
```diff
@@ -62,6 +62,7 @@ is.symbol = x => typeof x === 'symbol';
 
 is.array = Array.isArray;
 is.function = x => typeof x === 'function';
+is.class = x => typeof x === 'function' && /^\s*class\s+/.test(x.toString());
```
sindresorhus: `/^\s*class\s+/.test(x.toString());` => `x.toString().startsWith('class ');`

I don't think `.toString()` includes leading whitespace.

sindresorhus:
Also needs to be added to the readme.

SchneiderOr:
Okey, tests fixed and i've added description to the readme.

sindresorhus:
Thank you :)

SchneiderOr:
Sure, nice package :)

sudo-suhas:
A small note, this does not work if the class declaration was transpiled with babel.

PR #3, opened 2017-09-23
Add `is.inRange()`

glhrmv:


sudo-suhas:
Destructuring isn't supported on node 4.

sindresorhus:
Yes

sindresorhus:
Let’s call it `inRange`. And let’s do the Math.min/max thing you suggested so the range parameters can be specified in any order and also be negative.

glhrmv:
How's this?

brandon93s:
Could also borrow from [sindresorhus/in-range](https://github.com/sindresorhus/in-range/blob/master/index.js). It allows start to be optional, with a default of 0.

test.js
```diff
@@ -293,3 +293,10 @@ test('is.typedArray', t => {
 	t.false(m.typedArray([]));
 	t.false(m.typedArray({}));
 });
+
+test('is.inRange', t => {
+	const x = 3;
+	const arr = [0, 5];
```
sindresorhus: You can just put this inline in the assertion.

I'd also like to see assertions for `[5, 0]`, `[-5, 5]`, `[5, -5]`.

sindresorhus:
> Could also borrow from sindresorhus/in-range. It allows start to be optional, with a default of 0.

Good idea. I forgot I had that module. We could also support `is.inRange(5, 10)`. So the user can either use an array as range or a number as upper-bound. Thoughts?

brandon93s:
I agree. The flexibility of supporting both inputs would be nice.

test.js
gioragutt: Also add assertions for `is.inRange` that return `false`.

I also agree with inlining the test data.

glhrmv:
How about now? 😃

sindresorhus:
Perfect! Thanks for working on this @glhrmv 🙌

glhrmv:
@sindresorhus Thank you! It was my pleasure.

PR #4, opened 2017-09-24
Add `is.generator()` and `is.generatorFunction()`

yeskunall:
Took one thing off of the list! Feel free to merge. ❤️

PS: Yes, I notice there's some code duplication in my tests. Don't know how `ava` works in the slightest. If anyone knows how to group two tests into one to allow variable reuse, I'm more than willing to update the PR. 💪

### DIFF
--- a/index.js
+++ b/index.js
@@ -81,6 +81,10 @@ is.promise = x => {
 		);
 };
 
+is.generator = x => x && typeof x.next === 'function' && typeof x.throw === 'function';
+
+is.generatorFunction = x => x.constructor.name === 'GeneratorFunction';
+
 is.regExp = x => getObjectType(x) === 'RegExp';
 is.date = x => getObjectType(x) === 'Date';
 is.error = x => getObjectType(x) === 'Error';
--- a/index.js
+++ b/index.js
@@ -81,9 +81,9 @@ is.promise = x => {
 		);
 };
 
-is.generator = x => x && typeof x.next === 'function' && typeof x.throw === 'function';
+is.generator = x => is.iterable(x) && is.function(x.next) && is.function(x.throw);
 
-is.generatorFunction = x => x.constructor.name === 'GeneratorFunction';
+is.generatorFunction = x => getObjectType(x) === 'GeneratorFunction';
 
 is.regExp = x => getObjectType(x) === 'RegExp';
 is.date = x => getObjectType(x) === 'Date';
--- a/test.js
+++ b/test.js
@@ -10,7 +10,11 @@ const ErrorSubclassFixture = class extends Error {};
 const types = new Map([
 	['undefined', undefined],
 	['null', null],
-	['string', '🦄'],
+	['string', [
+		'🦄',
+		'hello world',
+		''
+	]],
 	['number', [
 		6,
 		1.4,
@@ -318,4 +322,39 @@ test('is.typedArray', t => {
 	t.false(m.typedArray(new ArrayBuffer(1)));
 	t.false(m.typedArray([]));
 	t.false(m.typedArray({}));
+});
+
+test('is.inRange', t => {
+	const x = 3;
+
+	t.true(m.inRange(x, [0, 5]));
+	t.true(m.inRange(x, [5, 0]));
+	t.true(m.inRange(x, [-5, 5]));
+	t.true(m.inRange(x, [5, -5]));
+	t.false(m.inRange(x, [4, 8]));
+	t.true(m.inRange(-7, [-5, -10]));
+	t.true(m.inRange(-5, [-5, -10]));
+	t.true(m.inRange(-10, [-5, -10]));
+
+	t.true(m.inRange(x, 10));
+	t.true(m.inRange(0, 0));
+	t.true(m.inRange(-2, -3));
+	t.false(m.inRange(x, 2));
+	t.false(m.inRange(-3, -2));
+
+	t.throws(() => {
+		m.inRange(0);
+	});
+
+	t.throws(() => {
+		m.inRange(0, []);
+	});
+
+	t.throws(() => {
+		m.inRange(0, [5]);
+	});
+
+	t.throws(() => {
+		m.inRange(0, [1, 2, 3]);
+	});
 });
--- a/test.js
+++ b/test.js
@@ -82,7 +82,7 @@ const types = new Map([
 	['integer', 6]
 ]);
 
-// This ensure a certain method matches only the types
+// This ensures a certain method matches only the types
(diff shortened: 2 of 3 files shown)


test.js
```diff
@@ -166,6 +166,20 @@ if (isNode8orHigher) {
 	});
 }
 
+test('is.generator', t => {
+	const genObj = (function * () {
+		yield 42;
+	})();
+	t.true(m.generator(genObj));
+});
```
sindresorhus: Use the `testType` type of testing, so it tests against all the other tests too. You probably need to exclude `function`.

index.js
```diff
@@ -81,6 +81,10 @@ is.promise = x => {
 		);
 };
 
+is.generator = x => x && typeof x.next === 'function' && typeof x.throw === 'function';
+
+is.generatorFunction = x => x.constructor.name === 'GeneratorFunction';
```
sindresorhus: This fail throw on anything other than functions. You need to do:

```js
typeof x === 'function' && x.constructor && x.constructor.name === 'GeneratorFunction';
```

sindresorhus: But would be simpler (and probably safer) to just do `getObjectType(x) === 'GeneratorFunction'`

index.js
```diff
@@ -81,6 +81,10 @@ is.promise = x => {
 		);
 };
 
+is.generator = x => x && typeof x.next === 'function' && typeof x.throw === 'function';
```
sindresorhus: Should be documented that this is just a naive check, like done with `is.promise`.

sindresorhus: We can actually make the check a little bit better:

```js
is.generator = x => x && typeof x[Symbol.iterator] === 'function' && typeof x.next === 'function' && typeof x.throw === 'function';
```

sindresorhus:
> PS: Yes, I notice there's some code duplication in my tests. Don't know how ava works in the slightest.

That's not a problem. Tests should be atomic.

yeskunall:
Thanks for the review. I'll ping you once I've made the changes 😄

index.js
gioragutt: You can use `is.iterable` here

yeskunall: No, by definition, all generators are iterators. See [here](https://stackoverflow.com/a/19660350) why just checking for an `iterable` isn't enough in this case.

test.js
yeskunall: Hey @sindresorhus 👋

Stuck here. I was going through your `types` map, and was hoping to just plug in a generator function that returns an object, but that doesn't seem to suffice. Am I missing something here?

For reference:

```diff
+ ['generator', (function* () { yield 42; })()],
```

... and calling it as so:

```js
test('is.generator', t => {
	testType(t, 'generator', ['function']);
});
```

... gives me:

```
Value: undefined

  Value is not `false`:

  undefined
```

... although I believe, it should return a generator object, and hence, pass the test. :thinking:

sindresorhus: It's because of the `x &&` part of the detection. If it's false, it returns itself instead of false, so `undefined` would return `undefined` instead of `false.`

yeskunall: Ah, I see `undefined &&` will always return `undefined`.

However, it's weird because in all my testing so far, calling a generator function returns a generator object, so I'm not sure where it's returning `undefined`, and therefore breaking the chain.

I'll look further into it tho. Thanks for the input so far. Appreciate it. :')

index.js
gioragutt: I meant that some of that check is already done by `is.iterable`, not all of it. F.e the symbol check

yeskunall:
@gioragutt I don't believe there's a need to use the existing function, but if @sindresorhus agrees to it as well, sure, I'll rewrite the check to use `is.function` 😄

index.js
sindresorhus: Yes, let's use `is.iterable` here.

sindresorhus:
> Use existing checks in is, f.e is.function for generator, checking that next and error are functions. Reuse whenever possible.

👍

index.js
yeskunall: @gioragutt @sindresorhus coolio. @gioragutt I thought you meant just check if the object is iterable. Whoopsie. 🙊

yeskunall:
@sindresorhus ping! This PR is ready to be merged once you approve my wording for documenting that `is.generator` is a naive check. Here's what I came up with, but maybe the sentence could be improved?
<hr />

`is.generator`
> Returns `true` for any object that implements its own `.next()` and `.throw()` methods and has a function definition for `Symbol.iterator`.

yeskunall:
Hey @sindresorhus, you seem to be busy. But anyway, on Node 4.x.x, the check `getObjectType` on a generator function returns `Function`, so do you suggest I go back to my original check, or should I provision for Node 4.x.x, check for which version the user is running, and do two different checks?

Here's why the Travis build is failing:

```shell
yeskunall:~/workspace $ nvm use 4
Now using node v4.2.1 (npm v2.14.7)
```

```node
yeskunall:~/workspace $ node
> const toString = Object.prototype.toString;
undefined
> const getObjectType = x => toString.call(x).slice(8, -1);
undefined
> getObjectType(function* () { yield 42 });
'Function'
> const gen = function* () { yield 42; };
undefined
> gen.constructor.name;
'GeneratorFunction'
> # 🤔
```

Whenever you find time to respond. ❤️

sindresorhus:
> the check getObjectType on a generator function returns Function, so do you suggest I go back to my original check, or should I provision for Node 4.x.x, check for which version the user is running, and do two different checks?

Let's just go back to the old check, but include a comment about switching to the `getObjectType` check when we target Node.js 6 or newer.

sindresorhus:
Can you also fix the merge conflict?

yeskunall:
I'll make the changes as requested. Heading to work now, you should be able to merge this PR by today. 👍

yeskunall:
Hey @gioragutt, all done. Feel free to merge when you can. 👍

index.js
```diff
@@ -82,6 +82,11 @@ const hasPromiseAPI = x =>
 
 is.promise = x => is.nativePromise(x) || hasPromiseAPI(x);
 
+is.generator = x => is.iterable(x) && is.function(x.next) && is.function(x.throw);
+
+// Change to use `isObjectOfType` once Node 4.x.x LTS is dropped
+is.generatorFunction = x => x.constructor.name === 'GeneratorFunction';
```
sindresorhus: My first comment in https://github.com/sindresorhus/is/pull/4#discussion_r140665751 still applies.

yeskunall: Yeah, I'm sorry about that. Yesterday was not just my day when it came to coding 🙊

PR #6, opened 2017-09-25
Refactor to reduce code repetition

### LINKED ISSUE #5: is there any reason for `is.sharedArrayBuffer` special condition?
I'm doing a pull request to remove code repetition, and the sharedArrayBuffer method bothers me. 

```typescript
is.sharedArrayBuffer = x => {
	try {
		return getObjectType(x) === 'SharedArrayBuffer';
	} catch (err) {
		return false;
	}
};
```

Part of the pull request is making a method called `isTypeOf` (WIP, may change after review),

which is:

```typescript
const getObjectType = x => toString.call(x).slice(8, -1);
const isTypeOf = type => x => getObjectType(x) === type;
```

and the usage looks like:

```typescript
is.float32Array = isTypeOf('Float32Array');
is.float64Array = isTypeOf('Float64Array');

is.arrayBuffer = isTypeOf('ArrayBuffer');
```

whereas currently there's a `getObjectType(x) === 'some type'` repetition throughout the code.

What's the reason for `sharedArrayBuffer`s special check?

gioragutt:
* slight styling changes
* use existing methods whenever possible (f.e `is.function` instead of `typeof x === 'function`)
* change all occurrences of `is.something = x => getObjectType(x) === 'Something)` to `is.something = isTypeOf('Something')` using a factory method
* removed unnecessary `try/catch` in `is.sharedArrayBuffer`.

closes #5

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,4 @@
 'use strict';
-
 const util = require('util');
 
 const toString = Object.prototype.toString;
--- a/index.js
+++ b/index.js
@@ -1,4 +1,7 @@
 'use strict';
+
+const util = require('util');
+
 const toString = Object.prototype.toString;
 const getObjectType = x => toString.call(x).slice(8, -1);
 const isOfType = type => x => typeof x === type; // eslint-disable-line valid-typeof
@@ -151,7 +154,7 @@ is.inRange = (x, range) => {
 		return x >= Math.min.apply(null, range) && x <= Math.max.apply(null, range);
 	}
 
-	throw new TypeError('Invalid range');
+	throw new TypeError(`Invalid range: ${util.inspect}`);
 };
 
 module.exports = is;
--- a/index.js
+++ b/index.js
@@ -1,7 +1,8 @@
 'use strict';
 const toString = Object.prototype.toString;
 const getObjectType = x => toString.call(x).slice(8, -1);
-const isOfType = type => x => getObjectType(x) === type;
+const isOfType = type => x => typeof x === type; // eslint-disable-line valid-typeof
+const isObjectOfType = type => x => getObjectType(x) === type;
 
 const is = value => {
 	if (value == null) { // eslint-disable-line no-eq-null, eqeqeq
@@ -30,7 +31,7 @@ const is = value => {
 		return 'symbol';
 	}
 
-	if (type === 'function') {
+	if (is.function(value)) {
 		return 'Function';
 	}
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -59,69 +60,55 @@ is.string = x => typeof x === 'string';
 is.number = x => typeof x === 'number';
 is.boolean = x => typeof x === 'boolean';
 is.symbol = x => typeof x === 'symbol';
+is.nullOrUndefined = x => is.null(x) || is.undefined(x);
```
sindresorhus: Don't change the order. The order of defined methods is deliberate and should be the same as in the docs.

test.js
```diff
@@ -247,9 +251,7 @@ test('is.primitive', t => {
 		Symbol('🦄')
 	];
 
-	for (const el of primitives) {
-		t.true(m.primitive(el));
-	}
+	primitives.forEach(el => t.true(m.primitive(el)));
```
sindresorhus: I prefer how it was here. It was more readable and more efficient.

test.js
```diff
@@ -320,13 +322,15 @@ test('is.inRange', t => {
...
+	t.throws(() =>
+		t.true(m.inRange(0))
+	);
 
-	t.throws(() => {
-		t.true(m.inRange(0, [5]));
-	});
+	t.throws(() =>
+		t.true(m.inRange(0, [5]))
+	);
```
sindresorhus: I don't want this change. It's intentionally a code-block. I only do inline arrow functions when the return value is actually used, for readability reasons.

sindresorhus: I do notice now that the `t.true` inside these `t.throws` statements are moot though. So you can remove those.

sindresorhus:
Thanks for the cleanup. This looks great :)

index.js
gioragutt: So why would then nullOrUndefined be separated from other primitive predicates?

I accept the review, but I would still like an explanation, from the user perspective, not the dev.

test.js
gioragutt: Accepted, will revert this

test.js
gioragutt: I'll fix this if needed, but I'd like to ask first.
Why keep the `{ }`? IMO what adds to the readability is that the *throwing statement* is between new lines.

IMO there's nothing wrong with not wrapping a `void` statement in `{ }`, as they're eventually "dead characters", they don't have any meaning unless they contain more than two statements, that's just my opinion.

My example would be the `do` operator in `RxJS.Observable`. 
Often you do `.do(value => console.log(value))`. If I'd wrap every simple statement, I'd end up with something like: `.do(value => { console.log(value); })` which just clutters the code.

I'll revert it if you insist, but that's just my opinion.

sindresorhus: The following makes it perfectly clear `bar()` doesn't return anything and that `foo()` doesn't expect it to:

```js
foo(x => {
  bar();
});
```

This here, however, makes it seem like `bar()` returns something and `foo()` uses that return value:

```js
foo(x => bar());
```

I prefer explicit and clear code over saving a couple of characters.

index.js
sindresorhus: Because it's not a pure type check, it's a combinatory check, and therefore is better placed in Misc. Same as the `primitive` and `typedArray` check. Happy to reconsider if you have some good arguments against that, though.

gioragutt: I do understand now, what I had in mind in the moment is that because these are `primitive` related methods, then they should be placed alongside all other `primitive` methods.

also, I wanted to use it in a method that comes before where `nullOrUndefined` is originally placed.

Same for `typedArray`, there are several specific checks for all the different `typedArray`s, so it'd be logical if it was placed next to it.

test.js
gioragutt: You know what, I do agree with you on this point. Will revert.

gioragutt:
@sindresorhus Hey, I fixed the reviews, you can check it out now

sindresorhus:
Great! Thanks for cleaning this up :)

Would you be interested in being added as a maintainer on this project? No worries if not.

gioragutt:
Yes I would👍

PR #11, opened 2017-10-03
Add is.domElement()

melvin0008:
Check if the value passed is a dom Element or not.

Reference: https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType

### DIFF
--- a/index.js
+++ b/index.js
@@ -161,11 +161,12 @@ is.inRange = (x, range) => {
 	throw new TypeError(`Invalid range: ${util.inspect(range)}`);
 };
 
+const NODE_TYPE_ELEMENT_NODE = 1;
+const DOM_PROPERTIES_TO_CHECK = ['innerHTML', 'ownerDocument', 'style', 'attributes', 'nodeValue'];
+
 is.domElement = x => {
-	const ELEMENT_NODE = 1;
-	const propsToCheck = ['innerHTML', 'ownerDocument', 'style', 'attributes', 'nodeValue'];
-	return is.object(x) && x.nodeType === ELEMENT_NODE && is.string(x.nodeName) &&
-				!is.plainObject(x) && propsToCheck.every(prop => prop in x);
+	return is.object(x) && x.nodeType === NODE_TYPE_ELEMENT_NODE && is.string(x.nodeName) &&
+				!is.plainObject(x) && DOM_PROPERTIES_TO_CHECK.every(property => property in x);
 };
 
 is.infinite = x => x === Infinity || x === -Infinity;
--- a/index.js
+++ b/index.js
@@ -161,4 +161,6 @@ is.inRange = (x, range) => {
 	throw new TypeError(`Invalid range: ${util.inspect(range)}`);
 };
 
+is.domElement = x => is.object(x) && x.nodeType === 1;
+
 module.exports = is;
--- a/index.js
+++ b/index.js
@@ -161,6 +161,6 @@ is.inRange = (x, range) => {
 	throw new TypeError(`Invalid range: ${util.inspect(range)}`);
 };
 
-is.domElement = x => is.object(x) && x.nodeType === 1;
+is.domElement = x => is.object(x) && x.nodeType === 1 && is.string(x.nodeName) && !is.plainObject(x);
 
 module.exports = is;
--- a/index.js
+++ b/index.js
@@ -161,13 +161,17 @@ is.inRange = (x, range) => {
 	throw new TypeError(`Invalid range: ${util.inspect(range)}`);
 };
 
-const NODE_TYPE_ELEMENT_NODE = 1;
-const DOM_PROPERTIES_TO_CHECK = ['innerHTML', 'ownerDocument', 'style', 'attributes', 'nodeValue'];
+const NODE_TYPE_ELEMENT = 1;
+const DOM_PROPERTIES_TO_CHECK = [
+	'innerHTML',
+	'ownerDocument',
+	'style',
+	'attributes',
+	'nodeValue'
+];
 
-is.domElement = x => {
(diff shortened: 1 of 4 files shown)


melvin0008:
Reference: https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType

readme.md
```diff
@@ -145,6 +145,10 @@ Check if `value` (number) is in the range of `0` to `upperBound`.
 is.inRange(3, 10);
 ```
 
+##### .domElement(value)
+
+Returns `true` if `value` is a domElement.
```
kevva: > Returns `true` if `value` is a DOM element.

readme.md
```diff
@@ -145,6 +145,10 @@ Check if `value` (number) is in the range of `0` to `upperBound`.
 is.inRange(3, 10);
 ```
 
+##### .domElement(value)
```
kevva: Maybe `.domNode`? Just a question.

melvin0008: There are 7 types of nodes supported currently .
Reference : https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType

Element is one of them. So I feel .domElement() is more appropriate for this pull request

However I create a different function .domNode() which will check if value is a node?

kevva: > Element is one of them. So I feel .domElement() is more appropriate for this pull request

Makes sense.

> However I can create a different function .domNode() which will check if value is a node or not?

I don't think its use case is that common, but I'll let @sindresorhus decide if he wants it.

index.js
```diff
@@ -161,4 +161,6 @@ is.inRange = (x, range) => {
 	throw new TypeError(`Invalid range: ${util.inspect(range)}`);
 };
 
+is.domElement = x => is.object(x) && x.nodeType === 1;
```
sindresorhus: I think this is a bit too loose of a check.

Not sure what else we should check, but see https://www.google.co.th/search?q=is+dom+element+javascript+check&oq=is+dom+element+javascript+check&aqs=chrome..69i57j0l2j69i64l3.4818j0j1&sourceid=chrome&ie=UTF-8 and https://github.com/lodash/lodash/blob/master/isElement.js for some ideas.

readme.md
sindresorhus: > I don't think its use case is that common, but I'll let @sindresorhus decide if he wants it.

Nah, only interested in `domElement`.

index.js
melvin0008: Yes, I do agree now. Any object with a nodeType would be selected as a domElement. I will change this behavior and update the pull request.

melvin0008:
I added additional checks to make sure we do not accept plain objects. 

However if a object looks like this 

```js
class Foo {
    get nodeType() {
	return 1;
  }
  get nodeName() {
	return 'div';
  }
}

is.domElement(new Foo())
```

it will return true. It also returns true in lodash.

sindresorhus:
Maybe we should check if a few more properties exist, just be sure. For example, these:

- `.className`
- `.style`
- `.ownerDocument`
- `.innerHTML`
- `.setAttribute`

test.js
```diff
@@ -358,3 +359,18 @@ test('is.inRange', t => {
...
+	t.true(m.domElement(document.createElement('div')));
+	t.false(m.domElement('hello world'));
+	t.false(m.domElement([]));
+	t.false(m.domElement(new Map()));
+	t.false(m.domElement(null));
+	t.false(m.domElement(undefined));
+	t.false(m.domElement(0));
+	t.false(m.domElement(NaN));
+	t.false(m.domElement(Infinity));
+	t.false(m.domElement({nodeType: 1, nodeName: 'div'}));
```
sindresorhus: Can you use the `testType` helper?

melvin0008: Yes sure. I'll update the review

index.js
```diff
@@ -161,4 +161,10 @@ is.inRange = (x, range) => {
 	throw new TypeError(`Invalid range: ${util.inspect(range)}`);
 };
 
+is.domElement = x => {
+	const propsToCheck = ['innerHTML', 'ownerDocument', 'style', 'attributes', 'nodeValue'];
+	return is.object(x) && x.nodeType === 1 && is.string(x.nodeName) &&
```
gioragutt: what does 1 stand for here? please convert `"magic numbers"` to consts.

test.js
```diff
@@ -89,7 +92,8 @@ const types = new Map([
 		Object.create(null),
 		new Object() // eslint-disable-line no-new-object
 	]],
-	['integer', 6]
+	['integer', 6],
+	['domElement', document.createElement('div')]
```
gioragutt: What about more than a div?

you can do `['div', 'input', 'span', 'img', 'canvas', 'h1', 'h2', 'h3' ...].map(e => document.createElement(e))`

sindresorhus: Agreed, but we don't need the different `h` tags.

gioragutt:
I had just given some examples as a guideline, there are plenty more
examples.

בתאריך 5 באוק׳ 2017 5:02 אחה״צ,‏ "Sindre Sorhus" <notifications@github.com>
כתב:

test.js
sindresorhus: Let's go with: `div`, `input`, `span`, `img`, `canvas`, `script`.

gioragutt: Also, what markup object is not considered a domElement? So that we can check for more cases

melvin0008:
@gioragutt Thanks for the review. Learnt few things there

gioragutt:
@melvin0008 glad to help dude, thanks for the PR!

@sindresorhus any reason not to merge this?

sindresorhus:
@gioragutt I reviewed it and then got distracted and forgot about it.

@melvin0008 Nice work! :)

melvin0008:
Thank You:)

PR #12, opened 2017-10-04
boolean can be detected with `typeof`

KhaledElAnsari:
according to mdn `typeof` return string boolean for the boolean type (`true` and `false`)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,7 +17,7 @@ const is = value => {
 		return 'undefined';
 	}
 	
-	if (value === 'boolean') {
+	if (type === 'boolean') {
 		return 'boolean';
 	}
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
I’m aware, but using true and false was intentional. It’s slightly faster.

gioragutt:
@sindresorhus if so, shouldn't `is.boolean` be `val => val === true || val === false`?

sindresorhus:
Yes

KhaledElAnsari:
good point @gioragutt

gioragutt:
@KhaledElAnsari mind changing that in a PR?

PR #14, opened 2017-10-05
Add is.infinite()

melvin0008:
Add is.infinite with tests. I feel is.infinte is better than is.infinity.

### DIFF
--- a/test.js
+++ b/test.js
@@ -358,3 +358,11 @@ test('is.inRange', t => {
 		m.inRange(0, [1, 2, 3]);
 	});
 });
+
+test('is.infinite', t => {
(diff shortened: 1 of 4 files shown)


test.js
```diff
@@ -358,3 +358,11 @@ test('is.inRange', t => {
 		m.inRange(0, [1, 2, 3]);
 	});
 });
+
+test('is.infinite', t => {
+	t.true(m.infinite(Infinity));
+	t.true(m.infinite(-Infinity));
+	t.false(m.infinite(NaN));
+	t.false(m.infinite(5));
+	t.false(m.infinite('hello world'));
```
sindresorhus: Use the `testType` helper instead.

PR #15, opened 2017-10-05
updating `is.boolean` method

KhaledElAnsari:
According to [this comment](https://github.com/sindresorhus/is/pull/12#issuecomment-334327363) comparing variable to boolean values `true` and `false` manually is faster than using `typeof` so I'm changing the function for a faster results

### DIFF
--- a/index.js
+++ b/index.js
@@ -61,7 +61,7 @@ is.undefined = isOfType('undefined');
 is.null = x => x === null;
 is.string = isOfType('string');
 is.number = isOfType('number');
-is.boolean = isOfType('boolean');
+is.boolean = (value) => value === true || value === false;
 is.symbol = isOfType('symbol');
 
 is.array = Array.isArray;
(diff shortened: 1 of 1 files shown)


gioragutt:
btw, @sindresorhus, do you have a source on that claim?,

also, @KhaledElAnsari, for consistency, remove the `( )` around `value`: `(value) => ...` => `value => ...`,
also, the CI failed, make sure to see what the failure report says.

gioragutt:
Thanks @KhaledElAnsari !

KhaledElAnsari:
no problem @gioragutt

gioragutt:
@sindresorhus got it, sorry. I believe we can enforce that in the CI, I know that `@angular/cli` does that.

sindresorhus:
@gioragutt No worries. I forgot to tell you. I don't want to enforce it as nobody follows that. I've tried many times before. Just easier for us to rename the PR and squash merge.

sindresorhus:
> btw, @sindresorhus, do you have a source on that claim?,

I benchmarked it a year ago and then it was faster.

gioragutt:
@sindresorhus cool. I'll know from now on.

Issue #1, opened 2017-09-21
Ideas


sindresorhus:
I'm gonna try hard not to go overboard with stuff I will never need, but just a quick brain dump:

- [x] `.withinRange(3, [0, 5])`
- [x] `.generator()`
- [x] `.domElement()`
- [x] `.generatorFunction()`
- [x] `.infinity()` (Would match `Infinity` and `-Infinity`) (Maybe `.infinite()` would be a better name?)
- [x] `.empty([])` (string, array, object, map, set, ...)
- [x] `.blob()`
- [x] `.asyncFunction()`
- [x] And `.any()`/`.all()` methods to match multiple types.

Feedback wanted! 😀

glhrmv:
How are you thinking `.withinRange` should work? Is something like this OK?
```js
is.withinRange = (x, arr) => (x > Math.min.apply(Math, arr)) && (x < Math.max.apply(Math, arr));
```

sindresorhus:
@glhrmv I was thinking:

```js
is.withinRange = (input, [lower, upper]) => input >= lower && input <= upper;
```

glhrmv:
Array destructuring is definitely more elegant, I forgot that was a thing.

As for `.domElement`, that'll require following [this procedure](https://github.com/avajs/ava/blob/master/docs/recipes/browser-testing.md) for testing, right? If so, I can get on that as well.

sindresorhus:
> As for .domElement, that'll require following this procedure for testing, right?

Yup. Let’s do it in a separate test file.

glhrmv:
I couldn't make `browser-env` (or even `window`) make `document` defined in the test file, for some reason. In any case, the idea I had of `is.domElement(x)` was to make sure `x` is an `instanceof Element`, but obviously Node doesn't support `Element`, either. 

I'm giving this up to anyone who wants to deal with it, and I'm interested to see how this type of issue is solved.

sindresorhus:
@glhrmv You can see how it's done here: https://github.com/sindresorhus/linkify-urls/blob/4db31ab79d62dc0adf6b3c086a6a68f955343918/test.js#L2-L7

yeskunall:
With #4:

* [x] `.generator()`
* [x] `.generatorFunction()`

nrebhun:
Hey @sindresorhus, Hacktoberfest is upon us! 😜 

How about getting some more eyes on this issue/repo with the relevant tag?

fega:
[...]

PR #16, opened 2017-10-05
Add is.empty()

kodie:
Added the `is.empty(value)` function as mentioned in #1.

### DIFF
--- a/index.js
+++ b/index.js
@@ -163,4 +163,12 @@ is.inRange = (x, range) => {
 
 is.infinite = x => x === Infinity || x === -Infinity;
 
+is.empty = value => {
+	return (
+		((is.string(value) || is.array(value)) && !value.length) ||
+		(!is.map(value) && !is.set(value) && is.object(value) && !Object.keys(value).length) ||
+		((is.map(value) || is.set(value)) && !value.size)
+	);
+};
+
 module.exports = is;
--- a/index.js
+++ b/index.js
@@ -163,12 +163,10 @@ is.inRange = (x, range) => {
 
 is.infinite = x => x === Infinity || x === -Infinity;
 
-is.empty = value => {
-	return (
-		((is.string(value) || is.array(value)) && !value.length) ||
-		(!is.map(value) && !is.set(value) && is.object(value) && !Object.keys(value).length) ||
-		((is.map(value) || is.set(value)) && !value.size)
-	);
-};
+const isEmptyStringOrArray = x => (is.string(x) || is.array(x)) && x.length === 0;
+const isEmptyObject = x => !is.map(x) && !is.set(x) && is.object(x) && Object.keys(x).length === 0;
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -163,4 +163,12 @@ is.inRange = (x, range) => {
 
 is.infinite = x => x === Infinity || x === -Infinity;
 
+is.empty = value => {
```
gioragutt: you've got quite a lot going on in this method.
1. get rid of unnecessary wrapping `( )`.
1. `value.length === 0` over `!value.length`
1. you basically have 3 assertions inside (aka `assertion1 || assertion2 || assertion3`). extract each assertion to a helper method (like `hasPromiseAPI`), and invoke them inside `is.empty`. This will also help identify which case each assertion relates to, and that will also help clarify which part of the method each assertion in the tests relates to.
1. what is the outcome of `is.empty(undefined)` or `is.empty(null)`? we need to make sure that all cases are taken care of, as the [README](https://github.com/sindresorhus/is#why-yet-another-type-checking-module) states,

readme.md
```diff
@@ -150,6 +150,11 @@ is.inRange(3, 10);
 Check if `value` is `Infinity` or `-Infinity`.
 
 
+##### .empty(value)
+
+Returns `true` if `value` is an empty string, array, object, map, or set.
```
gioragutt: Rephrase it to match all other methods:
> Check if `value` is an empty string, array, object, map or set

test.js
```diff
@@ -373,3 +373,26 @@ test('is.inRange', t => {
 test('is.infinite', t => {
 	testType(t, 'infinite', ['number']);
 });
+
+test('is.empty', t => {
```
gioragutt: add cases for `null` and `undefined`.

gioragutt:
@kodie nice PR, personally been waiting for this in my app. Couple of changes for readability and clean code, but it seems fine otherwise.

kodie:
@gioragutt Fixing those right now!

Quick question, what should booleans return?

```js
is.empty(true);
is.empty(false);
```

?

gioragutt:
@kodie not quite sure, I believe that `is.empty(boolean) === !boolean`, but that's a thought.

@sindresorhus your thought?

kodie:
@gioragutt I've made the requested changes. I made it just return true if given a falsy value. So `true` will return false but `false`, `null`, and `undefined` will all return true.

readme.md
```diff
@@ -152,7 +152,7 @@ Check if `value` is `Infinity` or `-Infinity`.
 
 ##### .empty(value)
 
-Returns `true` if `value` is an empty string, array, object, map, or set.
+Returns `true` if `value` is falsy or an empty string, array, object, map, or set.
```
gioragutt: Change to
> Check that `value` is ...

sindresorhus:
Looks good. Thank you @kodie. Keep 'em coming 👌😎

PR #17, opened 2017-10-05
Add link to is-blob

melvin0008:


gioragutt:
@sindresorhus looking at `is-blob`, it can be easily integrated into `is`, without adding it as a dependency

readme.md
```diff
@@ -177,6 +177,7 @@ The most common mistakes I noticed in these modules was using `instanceof` for t
 - [is-array-sorted](https://github.com/sindresorhus/is-array-sorted) - Check if an Array is sorted
 - [is-error-constructor](https://github.com/sindresorhus/is-error-constructor) - Check if a value is an error constructor
 - [is-empty-iterable](https://github.com/sindresorhus/is-empty-iterable) - Check if an Iterable is empty
+- [is-blob] (https://github.com/sindresorhus/is-blob) - Check if a value is a Blob - File-like object of immutable, raw data
```
gioragutt: `[is-blob] (...` -> `[is-block](...`. Notice the space

gioragutt: I have a typo in the review, but ignore it.

melvin0008:
Thanks for pointing that out. It totally missed me

melvin0008:
@gioragutt  I do agree is-blob can be integrated into is. Infact @sindresorhus included it in his list of ideas. Let me know if I should make a PR for it

sindresorhus:
I included it in my brain-dump, yes, but I don't think we should include it in `is` until we can find some actual use-cases for it. I've never seen anyone do a conditional check for a `Blob` before. I don't want `is` to end up as a junkyard of any kind of `is` method. We should focus on things most users will need.

PR #19, opened 2017-10-06
Add is.any() and is.all() methods

kodie:
Here is the `any()` and `all()` methods mentioned in #1.

I wasn't 100% sure how you envisioned it working but the way I did it was both methods accepts two arrays as parameters so that it's possible to check multiple variables for multiple types.

Let me know if you were thinking something different!

### DIFF
--- a/index.js
+++ b/index.js
@@ -169,8 +169,9 @@ const isEmptyMapOrSet = x => (is.map(x) || is.set(x)) && x.size === 0;
 
 is.empty = x => !x || isEmptyStringOrArray(x) || isEmptyObject(x) || isEmptyMapOrSet(x);
 
-is.any = (predicate, ...values) => {
+is.any = function (predicate, values) {
 	let ret = false;
+	values = Array.prototype.slice.call(arguments, 1);
 
 	values.forEach(value => {
 		if (predicate(value)) {
@@ -181,8 +182,9 @@ is.any = (predicate, ...values) => {
 	return ret;
 };
 
-is.all = (predicate, ...values) => {
+is.all = function (predicate, values) {
 	let ret = true;
+	values = Array.prototype.slice.call(arguments, 1);
 
 	values.forEach(value => {
 		if (!predicate(value)) {
--- a/index.js
+++ b/index.js
@@ -169,4 +169,121 @@ const isEmptyMapOrSet = x => (is.map(x) || is.set(x)) && x.size === 0;
 
 is.empty = x => !x || isEmptyStringOrArray(x) || isEmptyObject(x) || isEmptyMapOrSet(x);
 
+const isType = (value, type) => {
+	switch (String(type).toLowerCase()) {
+		case 'undefined':
+			return is.undefined(value);
+		case 'null':
+			return is.null(value);
+		case 'string':
+			return is.string(value);
+		case 'number':
+			return is.number(value);
+		case 'boolean':
+			return is.boolean(value);
+		case 'symbol':
+			return is.symbol(value);
+		case 'array':
+			return is.array(value);
+		case 'function':
+			return is.function(value);
+		case 'buffer':
+			return is.buffer(value);
+		case 'object':
+			return is.object(value);
+		case 'regexp':
+			return is.regExp(value);
+		case 'date':
+			return is.date(value);
+		case 'error':
+			return is.error(value);
+		case 'nativepromise':
+			return is.nativePromise(value);
+		case 'promise':
+			return is.promise(value);
+		case 'generator':
+			return is.generator(value);
+		case 'generatorfunction':
+			return is.generatorFunction(value);
+		case 'map':
+			return is.map(value);
+		case 'set':
+			return is.set(value);
+		case 'weakmap':
+			return is.weakMap(value);
+		case 'weakset':
+			return is.weakSet(value);
+		case 'int8array':
+			return is.int8Array(value);
+		case 'uint8array':
+			return is.uint8Array(value);
+		case 'uint8clampedarray':
+			return is.uint8ClampedArray(value);
+		case 'int16array':
+			return is.int16Array(value);
+		case 'uint16array':
+			return is.uint16Array(value);
+		case 'int32array':
+			return is.int32Array(value);
+		case 'uint32array':
+			return is.uint32Array(value);
+		case 'float32array':
+			return is.float32Array(value);
+		case 'float64arrat':
+			return is.float64Array(value);
+		case 'arraybuffer':
+			return is.arrayBuffer(value);
+		case 'sharedarraybuffer':
+			return is.sharedArrayBuffer(value);
+		case 'dataview':
+			return is.dataView(value);
+		case 'nan':
+			return is.nan(value);
+		case 'nullorundefined':
+			return is.nullOrUndefined(value);
+		case 'primitive':
+			return is.primitive(value);
+		case 'integer':
+			return is.integer(value);
+		case 'plainobject':
+			return is.plainObject(value);
+		case 'iterable':
+			return is.iterable(value);
+		case 'class':
+			return is.class(value);
+		case 'typedarray':
+			return is.typedArray(value);
+		default:
+			return false;
+	}
+};
+
+is.any = (values, types) => {
+	let ret = false;
+
+	values.forEach(value => {
+		types.forEach(type => {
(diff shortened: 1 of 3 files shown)


melvin0008:
@kodie At the bottom of the issue sindresorhus explained his idea of any and all. You can check it out. For eg. any(is.string, {}, true, '🦄');

kodie:
@melvin0008 oh shoot! Completely missed that. I will update my PR.

melvin0008:
Something on the lines of 

is.any = (func, ...values) => values.some(func);

is.all = (func, ...values) => values.every(func);

kodie:
PR Updated. 🍻

kodie:
ah, using rest parameters break Node v4 support. :/

kodie:
There we go 👍

index.js
```diff
@@ -169,4 +169,30 @@ const isEmptyMapOrSet = x => (is.map(x) || is.set(x)) && x.size === 0;
 
 is.empty = x => !x || isEmptyStringOrArray(x) || isEmptyObject(x) || isEmptyMapOrSet(x);
 
+is.any = function (predicate, values) {
```
gioragutt: 1. See [Array.prototype.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) for the predicate check.
1. Take the values as an array using `(predicate, ...values) => { ... }` instead of the slice.
1. Use arrow functions instead of an anonymous function. All other methods use this syntax. You have to be consistent with the rest of the code.

basically, I think the code can look like `is.any = (predicate, ...values) => values.some(predicate)`;
of course, you have to validate the `values` is an array using `is.array`.

index.js
```diff
@@ -169,4 +169,30 @@ const isEmptyMapOrSet = x => (is.map(x) || is.set(x)) && x.size === 0;
...
+	let ret = false;
+	values = Array.prototype.slice.call(arguments, 1);
```
gioragutt: Generally, just a note, don't use `arguments`, ES5 has given has enough ways to deprecate usage pf `arguments`.

If you really end up having to use `arguments`, make sure you document it well.

index.js
```diff
@@ -169,4 +169,30 @@ const isEmptyMapOrSet = x => (is.map(x) || is.set(x)) && x.size === 0;
...
+	values.forEach(value => {
+		if (predicate(value)) {
+			ret = true;
+		}
+	});
+
+	return ret;
+};
+
+is.all = function (predicate, values) {
```
gioragutt: [Array.prototype.every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every), same as before.

test.js
```diff
@@ -402,3 +402,17 @@ test('is.empty', t => {
 	tempSet.add(1);
 	t.false(m.empty(tempSet));
 });
+
+test('is.any', t => {
```
gioragutt: Missing test cases:
1. no items (aka `m.any(m.integer)`)
1. invalid predicate? what if the passed predicate is not a function? this should be validated

@sindresorhus say `is.function(predicate) === false`, do we throw an Error?

test.js
```diff
@@ -402,3 +402,17 @@ test('is.empty', t => {
...
+	t.true(m.any(m.string, {}, true, '🦄'));
+	t.true(m.any(m.object, false, {}, 'unicorns'));
+	t.false(m.any(m.boolean, '🦄', [], 3));
+	t.false(m.any(m.integer, true, 'lol', {}));
+});
+
+test('is.all', t => {
```
gioragutt: same as above

index.js
gioragutt: I now see that `...values` is impossible, So as I've said, put a comment stating the it's not possible due to node4 compatibility.

So keep the current logic, but use the array methods like before, and extract a common method to which you can pass the correct method (aka `Array.prototype.some` or `Array.prototype.every`).

index.js
kodie: What would I use instead of `arguments` if I can't use `...values`?

gioragutt: Notice the comment I've made above. I wrote it before I've went over the commits.

kodie: @gioragutt oh, gotcha!

kodie:
@gioragutt PR Updated again, let me know what you think.

index.js
```diff
@@ -169,4 +169,35 @@ const isEmptyMapOrSet = x => (is.map(x) || is.set(x)) && x.size === 0;
 
 is.empty = x => !x || isEmptyStringOrArray(x) || isEmptyObject(x) || isEmptyMapOrSet(x);
 
+const isAnyOrAll = (all, predicate, values) => {
```
gioragutt: Instead of passing `all` as a boolean, pass in a `method` param instead.

this would actually be either `Array.prototype.some` or `Array.prototype.every`,

and you'd call `method.call(values, predicate)`.

it makes sense more than passing a boolean. If we're speaking clean code and `SOLID`, this violates both `Single Responsibility Principle` and `Open Close Principle`, because it makes `isAnyOrAll` (which should be renamed, for the same reasons) responsible for both having the the calls to the methods inside of it, and needing the logic to decide which method is called.

By passing the method from the caller, we can allow more methods to use `isAnyOrAll`.

I'm sure that there are more method of `Array` that receive a predicate.

index.js
```diff
@@ -169,4 +169,35 @@ const isEmptyMapOrSet = x => (is.map(x) || is.set(x)) && x.size === 0;
...
+	// We have to do it this way to keep node v4 support.
+	// So here we convert it to an array and slice off the first item.
+	values = Array.prototype.slice.call(values, 1);
+
+	if (is.function(predicate) === false) {
+		throw new TypeError(`Invalid predicate: ${util.inspect(predicate)}`);
+	}
+
+	if (values.length === 0) {
+		throw new TypeError(`Invalid number of values: ${util.inspect(values.length)}`);
```
gioragutt: You can never really have a value other than `0`, so why call inspect on this? instead, make the error more readable, f.e: `No values passed to ${callerMethod}`, where `callerMethod` can be passed as an argument.

index.js
```diff
@@ -169,4 +169,35 @@ const isEmptyMapOrSet = x => (is.map(x) || is.set(x)) && x.size === 0;
...
+	if (all) {
+		return values.every(predicate);
+	}
+
+	return values.some(predicate);
+};
+
+// We have to use anonymous functions for the any() and all() methods
+// to get the arguments since we can't use rest parameters in node v4.
+is.any = function (predicate) {
```
gioragutt: `is.any = predicate => isAnyOrAll(...)`

kodie: Have to use an anonymous function in order to get the correct arguments object. For some reason the arguments object is weird with arrow functions.

index.js
kodie: Makes sense! What would you suggest we change the name of `isAnyOrAll` to?

gioragutt: predicateOnArray

index.js
gioragutt: God I hate old js.

melvin0008: https://stackoverflow.com/questions/30935336/official-information-on-arguments-in-es6-arrow-functions

Arrow functions dont have arguments.

kodie:
@gioragutt Updated.

index.js
gioragutt: Just for the sake of debugging, name the anonymous functions (isAll and isAny) so that they'll be named if they appear in a stack trace.

After that, I'll approve

gioragutt: Also, we now have `is.empty` for `length === 0`

sindresorhus: > Just for the sake of debugging, name the anonymous functions (isAll and isAny) so that they'll be named if they appear in a stack trace.

That is not necessary. JS engines are able to infer the name.

test.js
sindresorhus: > say is.function(predicate) === false, do we throw an Error?

Yes

index.js
sindresorhus: > Also, we now have is.empty for length === 0

I think it's overdoing it using `is.empty` for that.

sindresorhus:
👌😎

gioragutt:
@sindresorhus great! v0.3.0?

sindresorhus:
Already done an hour ago ;)

PR #23, opened 2017-10-17
Add is.even() and is.odd()

brandon93s:
Intentionally left `even` and `odd` out of `types` in the tests for these ones. Integer / number values are growing in their need to be explicitly excluded from one another. Glad to add them if that's preferred.

### DIFF
--- a/index.js
+++ b/index.js
@@ -178,6 +178,10 @@ is.domElement = x => is.object(x) && x.nodeType === NODE_TYPE_ELEMENT && is.stri
 
 is.infinite = x => x === Infinity || x === -Infinity;
 
+const isAbsoluteMod2 = value => x => is.integer(x) && Math.abs(x % 2) === value;
+is.even = isAbsoluteMod2(0);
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -178,6 +178,10 @@ is.domElement = x => is.object(x) && x.nodeType === NODE_TYPE_ELEMENT && is.stri
 
 is.infinite = x => x === Infinity || x === -Infinity;
 
+const isAbsoluteMod2 = value => x => is.number(x) && Math.abs(x % 2) === value;
```
sindresorhus: `is.number()` => `is.integer()`, since per docs it only works on integers anyway.

sindresorhus:
> Intentionally left even and odd out of types in the tests for these ones. Integer / number values are growing in their need to be explicitly excluded from one another. 

That's ok.

index.js
brandon93s: 👍 Updated

sindresorhus:
🥇

PR #24, opened 2017-10-19
Add `is.emoji()`

yeskunall:
:unicorn: 🎊 🎉
<br/>
**NOTE**: The regex from the emoji comes from [this](https://medium.com/reactnative/emojis-in-javascript-f693d0eb79fb) article.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@ const toString = Object.prototype.toString;
 const getObjectType = x => toString.call(x).slice(8, -1);
 const isOfType = type => x => typeof x === type; // eslint-disable-line valid-typeof
 const isObjectOfType = type => x => getObjectType(x) === type;
+const emojiRegex = /(?:[\u2700-\u27bf]|(?:\ud83c[\udde6-\uddff]){2}|[\ud800-\udbff][\udc00-\udfff]|[\u0023-\u0039]\ufe0f?\u20e3|\u3299|\u3297|\u303d|\u3030|\u24c2|\ud83c[\udd70-\udd71]|\ud83c[\udd7e-\udd7f]|\ud83c\udd8e|\ud83c[\udd91-\udd9a]|\ud83c[\udde6-\uddff]|[\ud83c[\ude01-\ude02]|\ud83c\ude1a|\ud83c\ude2f|[\ud83c[\ude32-\ude3a]|[\ud83c[\ude50-\ude51]|\u203c|\u2049|[\u25aa-\u25ab]|\u25b6|\u25c0|[\u25fb-\u25fe]|\u00a9|\u00ae|\u2122|\u2139|\ud83c\udc04|[\u2600-\u26FF]|\u2b05|\u2b06|\u2b07|\u2b1b|\u2b1c|\u2b50|\u2b55|\u231a|\u231b|\u2328|\u23cf|[\u23e9-\u23f3]|[\u23f8-\u23fa]|\ud83c\udccf|\u2934|\u2935|[\u2190-\u21ff])/;
 
 const is = value => {
 	if (value === null) {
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -182,6 +182,8 @@ Returns `true` if `value` is an odd integer.
 
 Returns `true` if `value` is falsy or an empty string, array, object, map, or set.
 
+##### .emoji(value)
```
brandon93s: I'd prefer to see this _below_ `.emptyOrWhitespace` in the docs to:

1. Keep `.empty` and `.emptyOrWhitespace` together since they are similar in intent
2. Keep the order consistent with the source

Examples and / or a description would also be beneficial.

index.js
```diff
@@ -190,6 +191,7 @@ const isEmptyMapOrSet = x => (is.map(x) || is.set(x)) && x.size === 0;
 
 is.empty = x => !x || isEmptyStringOrArray(x) || isEmptyObject(x) || isEmptyMapOrSet(x);
 is.emptyOrWhitespace = x => is.empty(x) || isWhiteSpaceString(x);
+is.emoji = x => is.string(x) && emojiRegex.test(x);
```
brandon93s: The name implies we are checking for a single emoji, but this currently returns true for any string that _contains_ an emoji:

```js
emojiRegex.test('this is a test 🙏')
// => true
```

Should we consider a different name? Or behavior? Something along the lines of `is.stringWithEmoji` maybe.

yeskunall:
Yeah, this wasn't very well thought 🙊 I'm glad you started a discussion
tho. I'm out right now, but I will gladly make the changes based on where
this discussion goes.
-- 

Kunall Banerjee

brandon93s:
[Here](https://jsfiddle.net/usmgfewh/1/) is a quick test of `emojiRegex` against 1300+ emoji. All are passing.

index.js
```diff
@@ -5,6 +5,7 @@ const toString = Object.prototype.toString;
 const getObjectType = x => toString.call(x).slice(8, -1);
 const isOfType = type => x => typeof x === type; // eslint-disable-line valid-typeof
 const isObjectOfType = type => x => getObjectType(x) === type;
+const emojiRegex = /(?:[\u2700-\u27bf]|(?:\ud83c[\udde6-\uddff]){2}|[\ud800-\udbff][\udc00-\udfff]|[\u0023-\u0039]\ufe0f?\u20e3|\u3299|\u3297|\u303d|\u3030|\u24c2|\ud83c[\udd70-\udd71]|\ud83c[\udd7e-\udd7f]|\ud83c\udd8e|\ud83c[\udd91-\udd9a]|\ud83c[\udde6-\uddff]|[\ud83c[\ude01-\ude02]|\ud83c\ude1a|\ud83c\ude2f|[\ud83c[\ude32-\ude3a]|[\ud83c[\ude50-\ude51]|\u203c|\u2049|[\u25aa-\u25ab]|\u25b6|\u25c0|[\u25fb-\u25fe]|\u00a9|\u00ae|\u2122|\u2139|\ud83c\udc04|[\u2600-\u26FF]|\u2b05|\u2b06|\u2b07|\u2b1b|\u2b1c|\u2b50|\u2b55|\u231a|\u231b|\u2328|\u23cf|[\u23e9-\u23f3]|[\u23f8-\u23fa]|\ud83c\udccf|\u2934|\u2935|[\u2190-\u21ff])/;
```
sindresorhus: What's the source of this regex? How do we know it's correct?

yeskunall: @sindresorhus see [this](https://github.com/lodash/lodash/blob/4.16.6/lodash.js#L242). However, if I'm right, that doesn't cover all cases, as noted in [this](https://medium.com/reactnative/emojis-in-javascript-f693d0eb79fb) article. My regex is taken directly from the previously mentioned article. I'll update the PR description for everyone to see.

index.js
yeskunall: `is.stringWithEmoji` could work 🤔 let's see what Sindre has to say. 👍

yeskunall:
@sindresorhus ping! Feel like you might have missed my response to your comments. Close if it's no longer needed. 👍

sindresorhus:
I've been thinking about this, and I don't think it's worth adding after all. It's not something I would ever use and it has a very narrow use-case.

To be completely accurate, we would have to use this regex: https://github.com/mathiasbynens/emoji-regex/blob/master/index.js Which is 7 KB.

Instead, I've done a tiny module for this if anyone needs it: https://github.com/sindresorhus/has-emoji

Sorry for the wasted work @yeskunall.

yeskunall:
Understandable, and yes, I was going to suggest using [sindresorhus/has-emoji](https://github.com/sindresorhus/has-emoji) if this didn't work out for whatever reason.

No worries. ❤️

Issue #9, opened 2017-09-27
Use TypeScript


gioragutt:
I think that re-writing the code in `typescript` could really help this library, for the following reasons -
1. This will make it easy to integrate into `typescript` based codebases
1. This will help with readability and maintenance of the code
1. This will allow us to use the newest `JS`/`TS` features, while keeping compatibility with older `Node` versions
 
Think about the `inRange` pull-request - we could not use `array destructuring` because this messed up compatibility with `Node v4`.  
If we write this in `typescript`, the compiled `js` will be compatible with older versions, while the source code will be written with every `JS` feature we see fit.

sindresorhus:
Sure. I would like to see #4 merged first though.

sindresorhus:
This can be implemented now.

gioragutt:
@sindresorhus there seems to be an (unsurprising) problem when declaring the `null`, `function ` and `class` methods, as they are reserved keywords.

There are two ways to expose typescript.

1. to re-write entirely with `typescript` (and have the `.d.ts` file an artifact of the compilation)
1. to expose a manual `.d.ts`

If we mean to expose a manual `.d.ts` file, it would look something like this:

```typescript
declare module "@sindresorhus/is" {
    export interface Is {
        (val: any): string;
        undefined: TypeQueryMethod;
        null: TypeQueryMethod;
        string: TypeQueryMethod;
        number: TypeQueryMethod;
        boolean: TypeQueryMethod;
        symbol: TypeQueryMethod;
    
        array: TypeQueryMethod;
        function: TypeQueryMethod;
        buffer: TypeQueryMethod;
        object: TypeQueryMethod;
    
        nativePromise: TypeQueryMethod;
        promise: TypeQueryMethod;
        regExp: TypeQueryMethod;
        date: TypeQueryMethod;
        error: TypeQueryMethod;
        map: TypeQueryMethod;
        set: TypeQueryMethod;
        weakMap: TypeQueryMethod;
        weakSet: TypeQueryMethod;
    
        int8Array: TypeQueryMethod;
        uint8Array: TypeQueryMethod;
        uint8ClampedArray: TypeQueryMethod;
        int16Array: TypeQueryMethod;
        uint16Array: TypeQueryMethod;
        int32Array: TypeQueryMethod;
        uint32Array: TypeQueryMethod;
        float32Array: TypeQueryMethod;
        float64Array: TypeQueryMethod;
        arrayBuffer: TypeQueryMethod;
[...]

PR #28, opened 2017-10-31
Refactor module in TypeScript

ltetzlaff:
According to #9 I setup the typescript compiler, added typings where appropriate, otherwise `any` and configured the ignores so that npm will contain `index.d.ts` and `index.js` and github `index.ts`.

Due to `index.js` being "volatile" compiled code I had to drop xo (sorry Sindre) because the style of the compiled code [can't be configured](https://stackoverflow.com/a/29013080/3857476)

Also some keywords are reserved (`null`, `function`, `class`) and needed to be merged into the declaration via `Object.assign()`, see [keywords issue](https://github.com/Microsoft/TypeScript/issues/2536#issuecomment-154382218).


index.ts
```diff
@@ -0,0 +1,237 @@
+const util = require('util');
+
+//const toString = Object.prototype.toString;
```
sindresorhus: ?

tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+    "compileOnSave": true,
+    "compilerOptions": {
+        "target": "es6",
```
sindresorhus: Just set it to `es5` and we can use any ES6 feature, regardless of having to support Node.js 4.

tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+        "moduleResolution": "node",
+        "strict": true,
+        "alwaysStrict": true,
+        "declaration": true,
+        "outDir": ".",
+        "noImplicitAny": true,
+        "suppressImplicitAnyIndexErrors": true
+    },
```
sindresorhus: We should create a source map file too, for better debugging experience.

tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "include": [
+        "index.ts"
+    ],
+    "exclude": [
+        "node_modules"
+    ]
```
sindresorhus: Excluding `node_modules` is already the default.

index.ts
```diff
@@ -0,0 +1,237 @@
...
+	export const any = function (predicate: any) {
+		return predicateOnArray(Array.prototype.some, predicate, arguments);
+	};
+
+	export const all = function (predicate: any) {
+		return predicateOnArray(Array.prototype.every, predicate, arguments);
+	};
+}
+
+// Some few keywords are reserved
```
sindresorhus: Can you link to the TS issue here in a comment?

index.ts
```diff
@@ -0,0 +1,237 @@
...
+	export const domElement = (x: any) => is.object(x) && x.nodeType === NODE_TYPE_ELEMENT && is.string(x.nodeName) &&
+		!is.plainObject(x) && DOM_PROPERTIES_TO_CHECK.every(property => property in x);
+
+	export const infinite = (x: any) => x === Infinity || x === -Infinity;
+
+	const isAbsoluteMod2 = (value: number) => (x: number) => is.integer(x) && Math.abs(x % 2) === value;
+	export const even = isAbsoluteMod2(0);
+	export const odd = isAbsoluteMod2(1);
+
+	const isWhiteSpaceString = (x: any) => is.string(x) && /\S/.test(x) === false;
```
sindresorhus: Let's use a better name than `x`, since it will show up in auto-completion. We can use `value` instead.

index.ts
```diff
@@ -0,0 +1,237 @@
...
+	const isValidLength = (x: any) => is.safeInteger(x) && x > -1;
+	export const arrayLike = (x: any) => !is.nullOrUndefined(x) && !is.func(x) && isValidLength(x.length);
+
+	export const inRange = (x: number, range: number | number[]) => {
+		if (is.number(range)) {
+			return x >= Math.min(0, range as number) && x <= Math.max(range as number, 0);
+		}
+
+		if (is.array(range) && range.length === 2) {
+			// TODO: Use spread operator here when targeting Node.js 6 or higher
```
sindresorhus: When you target `es5` in the tsconfig, you can fix this TODO.

.npmignore
```diff
@@ -0,0 +1 @@
+index.ts
```
sindresorhus: This is not needed as we're using the `files` property in package.json

sindresorhus:
> Due to index.js being "volatile" compiled code I had to drop xo (sorry Sindre) because the style of the compiled code can't be configured

No worries

sindresorhus:
Thanks for working on this @ltetzlaff :)

sindresorhus:
Does it really make sense to export each method as an individual export? I'm just asking. Because without the `is` word, the readability suffers.

```js
import {string} from 'is';

...

string('🦄');
```

vs

```js
import is from 'is';

...

is.string('🦄');
```

You could, of course, do this, but I doubt most will:

```js
import {string as isString} from 'is';

...

isString('🦄');
```

sindresorhus:
> Due to index.js being "volatile" compiled code I had to drop xo (sorry Sindre) because the style of the compiled code can't be configured

I published a TSLint config which we can use.

Just run `npm i --save-dev tslint tslint-xo` and add a `tslint.json` file to the repo root with:

```json
{
     "extends": "tslint-xo"
}
```

And the following to `scripts` in package.json:

```json
"lint": "tslint --project tsconfig.json",
```

tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "compileOnSave": true,
+    "compilerOptions": {
+        "target": "es6",
+        "moduleResolution": "node",
+        "strict": true,
+        "alwaysStrict": true,
+        "declaration": true,
+        "outDir": ".",
+        "noImplicitAny": true,
+        "suppressImplicitAnyIndexErrors": true
```
sindresorhus: Why this?

tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+    "compileOnSave": true,
+    "compilerOptions": {
+        "target": "es6",
+        "moduleResolution": "node",
+        "strict": true,
+        "alwaysStrict": true,
+        "declaration": true,
+        "outDir": ".",
+        "noImplicitAny": true,
```
sindresorhus: This is not needed. It's included in the `strict` option.

index.ts
```diff
@@ -0,0 +1,237 @@
...
+	export const float64Array = isObjectOfType('Float64Array');
+
+	export const arrayBuffer = isObjectOfType('ArrayBuffer');
+	export const sharedArrayBuffer = isObjectOfType('SharedArrayBuffer');
+
+	export const truthy = (x: any) => !!x; // eslint-disable-line no-implicit-coercion
+	export const falsy = (x: any) => !x;
+
+	// the tests are designed for Number.isNaN rather than global isNaN
+	export const nan = (x: any) => is.number(x) && isNaN(Number(x));
```
sindresorhus: Why did you change this from the original?

index.ts
```diff
@@ -0,0 +1,237 @@
+const util = require('util');
```
SamVerschueren: `import * as util from 'util';`

index.ts
```diff
@@ -0,0 +1,237 @@
...
+
+//const toString = Object.prototype.toString;
+const getObjectType = (x: any) => toString.call(x).slice(8, -1) as string;
+const isOfType = (type: string) => (x: any) => typeof x === type; // eslint-disable-line valid-typeof
```
SamVerschueren: Remove `eslint-disable-line` comment

index.ts
```diff
@@ -0,0 +1,237 @@
...
+const isObjectOfType = (type: string) => (x: any) => getObjectType(x) === type;
+
+//export default is
+function is (value: any) {
```
SamVerschueren: Remove space before function paren

index.ts
```diff
@@ -0,0 +1,237 @@
...
+	export const uint16Array = isObjectOfType('Uint16Array');
+	export const int32Array = isObjectOfType('Int32Array');
+	export const uint32Array = isObjectOfType('Uint32Array');
+	export const float32Array = isObjectOfType('Float32Array');
+	export const float64Array = isObjectOfType('Float64Array');
+
+	export const arrayBuffer = isObjectOfType('ArrayBuffer');
+	export const sharedArrayBuffer = isObjectOfType('SharedArrayBuffer');
+
+	export const truthy = (x: any) => !!x; // eslint-disable-line no-implicit-coercion
```
SamVerschueren: Remove `eslint-disable-line` comment.

SamVerschueren: @sindresorhus Would it be better to use `Boolean(x)` here?

index.ts
```diff
@@ -0,0 +1,237 @@
...
+	]);
+	export const primitive = (x: any) => is.null_(x) || primitiveTypes.has(typeof x);
+
+	export const integer = Number.isInteger;
+	export const safeInteger = Number.isSafeInteger;
+
+	export const plainObject = (x: any) => {
+		// From: https://github.com/sindresorhus/is-plain-obj/blob/master/index.js
+		let prototype;
+		// eslint-disable-next-line no-return-assign
```
SamVerschueren: Remove `eslint-disable-next-line` comment.

index.ts
```diff
@@ -0,0 +1,237 @@
...
+			throw new TypeError(`Invalid number of values`);
+		}
+
+		return method.call(values, predicate);
+	};
+
+	// We have to use anonymous functions for the any() and all() methods
+	// to get the arguments since we can't use rest parameters in node v4.
+	export const any = function (predicate: any) {
+		return predicateOnArray(Array.prototype.some, predicate, arguments);
```
SamVerschueren: TS supports rest parameters.

index.ts
```diff
@@ -0,0 +1,237 @@
...
+	};
+
+	export const all = function (predicate: any) {
+		return predicateOnArray(Array.prototype.every, predicate, arguments);
```
SamVerschueren: Same as above, you can use rest parameters.

tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+        "suppressImplicitAnyIndexErrors": true
+    },
+    "include": [
+        "index.ts"
+    ],
+    "exclude": [
+        "node_modules"
+    ]
+}
```
SamVerschueren: Add final newline

index.ts
sindresorhus: Yes

index.ts
ltetzlaff: Number.isNaN is currently not supported and isNaN() is typeguarded to only accept number
see https://github.com/Microsoft/TypeScript/issues/15149

index.ts
ltetzlaff: for all methods?

index.ts
sindresorhus: Ok. Maybe you can use `// @ts-ignore` on it? https://blogs.msdn.microsoft.com/typescript/2017/10/31/announcing-typescript-2-6/

If not, at least link to the TS issue in a comment above.

index.ts
sindresorhus: Yes

tsconfig.json
ltetzlaff: The problem is with `new Set()`, which is perfectly used but typescript won't provide a polyfill for es5 during the transpilation, so I would suggest to ponyfill with array and `Array.includes` or an associative Object? 😦

sindresorhus: Should work if you specify:

```json
"lib": [
	"es2015"
],
```

ltetzlaff:
Hi guys, thanks a lot for pointing out those changes, I'm really glad to contribute here.
I've covered the suggestions and refactored the tests file.

Now there are several points that introduce some difficulties, for instance:

## is.class
	
Fails on *classes* because TS-es5 makes *functions* out of it.
	
<details>
 <summary>Example</summary>
	
```js
var Foo = /** @class */ (function () {
		function Foo() {
		}
		return Foo;
}());
var classDeclarations = [
		Foo,
		/** @class */ (function (_super) {
				__extends(Bar, _super);
				function Bar() {
						return _super !== null && _super.apply(this, arguments) || this;
				}
				return Bar;
		}(Foo))
];
```
</details>

## is.plainObject

Fails on *generators* because TS-es5 makes *plainObjects* out of it.

<details>
 <summary>Example</summary>
	
```js
var __generator = (this && this.__generator) || function (thisArg, body) {
	var _ = { label: 0, sent: function() { if (t[0] & 1) throw t[1]; return t[1]; }, trys: [], ops: [] }, f, y, t, g;
	return g = { next: verb(0), "throw": verb(1), "return": verb(2) }, typeof Symbol === "function" && (g[Symbol.iterator] = function() { return this; }), g;
	function verb(n) { return function (v) { return step([n, v]); }; }
	function step(op) {
...
```
</details>


## is.generatorFunction

Fails as well.

<details>
 <summary>Example</summary>
	
```js
return __generator(this, function (_a) {
		switch (_a.label) {
				case 0: return [4 /*yield*/, 4];
				case 1:
						_a.sent();
						return [2 /*return*/];
		}
});
```
</details>

.gitignore
```diff
@@ -1,2 +1,4 @@
 node_modules
 yarn.lock
+dist/**/*.js*
+dist/**/*.d.ts
```
sindresorhus: I would just ignore the whole directory instead:

```
dist
```

tests/test.ts
```diff
@@ -367,20 +406,20 @@ test('is.class', t => {
...
 	const typedArrays = [
-		new Int8Array(),
-		new Uint8Array(),
-		new Uint8ClampedArray(),
-		new Uint16Array(),
-		new Int32Array(),
-		new Uint32Array(),
-		new Float32Array(),
-		new Float64Array()
+		new Int8Array(0),
```
sindresorhus: These can be created without a value, but I assume TS does not agree with that? If so, could you open a TS issue an link to it here?

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Int8Array

tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+        "declaration": true,
+        "sourceMap": true,
+        "outDir": "dist",
+        "lib": [
+            "es2015", "dom", "scripthost"
+        ]
+    },
+    "include": [
+        "src/**/*.ts",
+        "tests/**/*.ts"
```
sindresorhus: I don't think we need this. By default, it will include all `.ts` files.

sindresorhus:
> Now there are several points that introduce some difficulties, for instance:

Hmm, yeah. TS gives very little individual control of what should be transpiled. I think the easiest would be to just use target `es2015` then. We can still use `Set` as it's supported in Node.js 4.

package.json
```diff
@@ -16,7 +16,7 @@
     "node": ">=4"
   },
   "scripts": {
```
sindresorhus: You have to add a `main` field to package.json with the `dist`.

sindresorhus: And update the `files` property.

src/index.ts
```diff
@@ -0,0 +1,234 @@
...
+	return 'Object';
+};
+
+namespace is {
+	export const undefined = isOfType('undefined');
+	export const string = isOfType('string');
+	export const number = isOfType('number');
+	export const boolean = (value: any) => value === true || value === false;
+	export const symbol = isOfType('symbol');
+	export const func = isOfType('function');
```
sindresorhus: Use `function_` instead of `func`, for consistency.

src/index.ts
```diff
@@ -0,0 +1,234 @@
...
+		'undefined',
+		'string',
+		'number',
+		'boolean',
+		'symbol'
+	]);
+
+	export const primitive = (value: any) => is.null_(value) || primitiveTypes.has(typeof value);
+
+	export const integer = (value: any) => is.number(value) && isFinite(value) && (value | 0) === value;
```
sindresorhus: Same here. Comment on why this is not using `Number.isInteger`

tests/test.ts
ltetzlaff: (added)]https://github.com/Microsoft/TypeScript/issues/19680], will add a comment in the code as well

ltetzlaff:
> I think the easiest would be to just use target es2015 then.

Alright, cool! That solves some of the other issues regarding static methods available through `Number`. 🎉

ltetzlaff:
Yay, next round:

## Node 4 Build fails:
<details>
 <summary>Destructuring</summary>

```
/home/travis/build/sindresorhus/is/node_modules/jsdom/lib/api.js:10
const { URL } = require("whatwg-url");
      ^
SyntaxError: Unexpected token {
```
</details>


## Node 8 Build fails:
<details>
 <summary>Typescript builds async functions to return this:</summary>

```
var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
    return new (P || (P = Promise))(function (resolve, reject) {
        function fulfilled(value) { try { step(generator.next(value)); } catch (e) { reject(e); } }
        function rejected(value) { try { step(generator["throw"](value)); } catch (e) { reject(e); } }
        function step(result) { result.done ? resolve(result.value) : new P(function (resolve) { resolve(result.value); }).then(fulfilled, rejected); }
        step((generator = generator.apply(thisArg, _arguments || [])).next());
    });
};```
</details>

File: .gitignore [hunk 26]
```diff
@@ -1,2 +1,3 @@
 node_modules
 yarn.lock
+dist/
```
SamVerschueren: Add final newline. Just use `dist` without `/`.

I recommend installing an [editorconfig](http://editorconfig.org/) plugin for your text editor. Things like final newlines will then be added automatically.

package.json
```diff
@@ -16,10 +16,15 @@
...
-    "test": "xo && ava"
+    "lint": "tslint --project .",
+    "build": "tsc",
+    "test": "npm run lint && npm run build && ava dist/tests",
+    "prepublish": "npm run build"
   },
   "files": [
```
SamVerschueren: Just set it to `dist`.

tslint.json
```diff
@@ -0,0 +1,3 @@
+{
+	"extends": "tslint-xo"
+}
```
SamVerschueren: Final newline

src/index.ts
```diff
@@ -0,0 +1,231 @@
+import * as util from 'util';
```
SamVerschueren: Rename the directory from `src` to `source`. @sindresorhus and I agreed upon this name in another project.

tests/test.ts
```diff
@@ -1,147 +1,187 @@
-import util from 'util';
-import test from 'ava';
-import {jsdom} from 'jsdom';
-import m from '.';
+import * as util from 'util';
```
SamVerschueren: I'd rather put the `tests` directory into the `source` directory. This way, everything regarding code sits together. This also means that the `index.ts` file is compiled to `dist/index.js` which seems a little better than `dist/src/index.js`.

The only thing that should be done then is in the `prepublish` hook. Install [`del-cli`](https://github.com/sindresorhus/del-cli) as dev dependency and set this as prepublish script.
```
npm run build && del dist/tests
```

tests/test.ts
```diff
@@ -1,147 +1,187 @@
...
+			document.createDocumentFragment()
+		]}
+	], ['infinite',
+		{is:  m.infinite, fixtures: [Infinity, -Infinity]}
+	]
+]);
 
-const document = jsdom();
-const createDomElement = el => document.createElement(el);
+/* This ensures a certain method matches only the types
```
SamVerschueren: Make comment single line

tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+    "compileOnSave": true,
+    "compilerOptions": {
+        /* es2015 requires moduleResolution and module to be set, 
+        see https://github.com/Microsoft/TypeScript/issues/8189 */
+        "target": "es2015",
```
SamVerschueren: @sindresorhus Might have missed this, but didn't we want to target `ES5`?

sindresorhus: @SamVerschueren Can't: https://github.com/sindresorhus/is/pull/28#issuecomment-341303651

sindresorhus:
> Should we drop jsdom and use something that supports Node 4? Do you maybe have anything specific in mind @sindresorhus ?

Don't upgrade jsdom. The existing version was outdated exactly for this reason, to keep Node.js 4 support.

sindresorhus:
> Should we drop the test or add an OR-condition that matches the typescript-generated?

Just comment out the test until we can figure out a solution for it.

tests/test.ts
sindresorhus: Agreed

sindresorhus:
@ltetzlaff Any thoughts on https://github.com/sindresorhus/is/pull/28#issuecomment-341026374 ?

src/index.ts
```diff
@@ -0,0 +1,231 @@
...
+
+	const hasPromiseAPI = (value: any) =>
+		!null_(value) &&
+		isObject(value) &&
+		function_(value.then) &&
+		function_(value.catch);
+
+	export const promise = (value: any) => nativePromise(value) || hasPromiseAPI(value);
+
+	// TODO: Change to use `isObjectOfType` once Node.js 6 or higher is targeted
```
gioragutt: Can this be used now?

tests/test.ts
```diff
@@ -1,147 +1,187 @@
...
+
+const document = new jsdom.JSDOM().window.document;
+const createDomElement = (el: string) => document.createElement(el);
+
+interface Test {is(value: any): boolean; fixtures: any[]; }
+const types = new Map<string, Test>([
+	['undefined',
+		{is: m.undefined, fixtures: [undefined]}
+	], ['null',
+		{is: m.null_, fixtures: [null]}
```
gioragutt: Why not `m.null`? Isn't the whole purpose of Lines `225-229` in the source file to allow that?

tests/test.ts
```diff
@@ -1,147 +1,187 @@
...
-const PromiseSubclassFixture = class extends Promise {};
-const ErrorSubclassFixture = class extends Error {};
+/* Currently out of order, see https://github.com/Microsoft/TypeScript/issues/15202
+class PromiseSubclassFixture<T> extends Promise<T> {}*/
+class ErrorSubclassFixture extends Error {}
+
+const document = new jsdom.JSDOM().window.document;
+const createDomElement = (el: string) => document.createElement(el);
+
+interface Test {is(value: any): boolean; fixtures: any[]; }
```
gioragutt: spread this to
```typescript
interface Test {
    is: (value: any) => boolean;
    fixtures: any[];
}
```

and add a new-line after that

tests/test.ts
```diff
@@ -1,147 +1,187 @@
...
+			-0,
+			Infinity,
+			-Infinity
+		]}
+	], ['boolean',
+		{is: m.boolean, fixtures: [true, false]}
+	], ['symbol',
+		{is: m.symbol, fixtures: [Symbol('🦄')]}
+	], ['array',
+		{is: m.array, fixtures: [[1, 2], new Array(2)]} // tslint:disable-line:prefer-array-literal
```
gioragutt: There are multiple fixtures and this is a single line.
I'm fine with single-lining if there's only one fixture, although I'd rather consistency overall,
Meaning that each type would look like:
```typescript
['typename',
    { is: m.typename, fixtures: [
        // there can be one or more fixtures here, but this would be consistent, as there are currently
        // tests with one fixture which are one lined, but also some that're multiple fixtures but also one lined
    ]}
]
```

tests/test.ts
```diff
@@ -188,7 +228,15 @@ test('is.buffer', t => {
 });
 
 test('is.object', t => {
-	for (const el of types.get('object')) {
+	const testData = types.get('object');
+
+	if (testData === undefined) {
+		t.fail('is.object not defined');
+
```
gioragutt: remove this empty line

.gitignore
ltetzlaff: Yea, thanks for the hint, was just about time to give my dotfiles a refresh 👍

ltetzlaff:
> Don't upgrade jsdom. The existing version was outdated exactly for this reason, to keep Node.js 4 support.

Alright, I updated because I did not find typings for that version, but going after the date I found that package version `jsdom@9.12.0` goes together with `@types/jsdom@2.0.31`. which in turn reenables Node 4 🎉

tests/test.ts
ltetzlaff: Agreed, makes perfect sense

tests/test.ts
ltetzlaff: It does work when using it in js but Typescript can't find a type definition for that due to it being added via defineProperties 😞 
I know it is not optimal, do you have a better idea or a method that merges the namespace with reserved keywords here? 
Unfortunately `@ts-ignore` did not work out in that case.

tests/test.ts
ltetzlaff: The linter config insists on that empty line before return, sry 👍

ltetzlaff:
> @ltetzlaff Any thoughts on #28 (comment) ?

Yes, I agree, `is` makes sense as a module so we should not export every member of the namespace.

I'd go for enabling both

```js
import is from 'is';
```

as well as

```js
import {is} from 'is';
```

sindresorhus:
> I'd go for enabling both

Why? I prefer one way to do something.

ltetzlaff:
> Why? I prefer one way to do something.

Since there is only one export shall we go for default export then?
```js
import is from 'is';
```

src/index.ts
ltetzlaff: no, the comment is still valid because the restricting factor is node runtime here

sindresorhus:
> Since there is only one export shall we go for default export then?

Yes

package.json
```diff
@@ -16,10 +16,14 @@
     "node": ">=4"
   },
   "scripts": {
-    "test": "xo && ava"
+    "lint": "tslint --project .",
+    "build": "tsc",
+    "test": "npm run lint && npm run build && ava dist/tests",
+    "prepublish": "npm run build && del dist/tests"
   },
+  "main": "dist",
```
SamVerschueren: Add `"typings": "dist/index.d.ts"` here as well. Not sure if needed in the `main` field  though, but I always point to the `index.js` file explicitely.

```json
"main": "dist/index.js",
"typings": "dist/index.d.ts",
```

source/tests/test.ts
```diff
@@ -1,147 +1,246 @@
-import util from 'util';
-import test from 'ava';
+import * as util from 'util';
+import test, {TestContext, Context} from 'ava';
 import {jsdom} from 'jsdom';
-import m from '.';
+import m from '..'; // tslint:disable-line:import-name
```
SamVerschueren: The rule is disabled in `tslint-xo@0.2.1` so you can remove the comment.

source/tests/test.ts
```diff
@@ -1,147 +1,246 @@
...
-	['infinite', [
-		Infinity,
-		-Infinity
-	]]
+interface Test {
+	is(value: any): boolean;
+	fixtures: any[];
+}
+
+const types = new Map<string, Test>([
```
SamVerschueren: In terms of readability, I would rather see it formatted like this

```ts
const types = new Map<string, Test>([
	['undefined', {
		is: m.undefined,
		fixtures: [
			undefined
		]
	}],
	['null', {
		is: m.null_,
		fixtures: [
			null
		]
	}],
	['string', {
		is: m.string,
		fixtures: [
			'🦄',
			'hello world',
			''
		]
	}]
	// ...
]);
```

source/tests/test.ts
```diff
@@ -1,147 +1,246 @@
...
+			-Infinity
+		]}
+	]
 ]);
 
-// This ensures a certain method matches only the types
-// it's supposed to and none of the other methods' types
-const testType = (t, type, exclude) => {
-	for (const [key, value] of types) {
+/* This ensures a certain method matches only the types
```
SamVerschueren: One-line comment

source/tests/test.ts
```diff
@@ -1,147 +1,246 @@
...
+
+	if (testData === undefined) {
+		t.fail(`is.${type} not defined`);
+
+		return;
+	}
+
+	const {is} = testData;
+
+	types.forEach(({fixtures}, key) => {
```
SamVerschueren: Use a `for..of` instead (like it was).

```ts
for (const [key, {fixtures}] of types) {
	if (exclude && exclude.indexOf(key) !== -1) {
		continue;
	}

	const assert = key === type ? t.true.bind(t) : t.false.bind(t);

	for (const fixture of fixtures) {
		assert(is(fixture), `Value: ${util.inspect(fixture)}`);
	}
}
```

source/tests/test.ts
gioragutt: Agreed

source/tests/test.ts
gioragutt: Also, there's the comment about the exclude, we can use `types.filter` to get rid of it

gioragutt: Also, there's the comment about the exclude, we can use `types.filter` to get rid of it

ltetzlaff: not unless we do `Array.from(types).filter()` since `types` is a Map, but feel free to do so, If you feel that that would be an improvement :v:

source/tests/test.ts
```diff
@@ -17,206 +17,243 @@ interface Test {
 }
 
 const types = new Map<string, Test>([
-	['undefined',
-		{is: m.undefined, fixtures: [
+	['undefined', {
+		is: m.undefined, fixtures: [
```
SamVerschueren: Put `fixtures` on a newline as well.

package.json
```diff
@@ -16,10 +16,15 @@
     "node": ">=4"
   },
   "scripts": {
-    "test": "xo && ava"
+    "lint": "tslint --project .",
+    "build": "tsc",
+    "test": "npm run lint && npm run build && ava dist/tests",
+    "prepublish": "npm run build && del dist/tests"
   },
+  "main": "dist/index",
```
SamVerschueren: `dist/index.js`

package.json
```diff
@@ -16,10 +16,15 @@
...
+  "types": "dist/index",
```
SamVerschueren: `dist/index.d.ts`

ltetzlaff:
What's your preferred way of distributing this library?
- distributing ts and compiling locally
- distributing only pure js, declaration and source map
- npm vs gh?

SamVerschueren:
distributing only pure js, declaration and source map

sindresorhus:
Landed. Thanks for all the hard work you put into this @ltetzlaff. We really appreciate it 🙌

superhighfive

gioragutt:
Amazing work @ltetzlaff, seriously.

ltetzlaff:
Thanks for your words, I really enjoyed contributing to you guys. 🎉 😃

PR #29, opened 2017-11-06
Restore ability to use default export in CommonJS

sindresorhus:
So you can use `require('is')` instead of `require('is').default`.

It's a hack, but it seems to work fine.

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -247,3 +247,7 @@ Object.defineProperties(is, {
 });
 
 export default is; // tslint:disable-line:no-default-export
+
+// For CommonJS default export support
+module.exports = is;
+module.exports.default = is;


sindresorhus:
// @ltetzlaff

SamVerschueren:
Sweet, never thought about this :)!

source/index.ts
```diff
@@ -247,3 +247,7 @@ Object.defineProperties(is, {
 });
 
 export default is; // tslint:disable-line:no-default-export
+
+// For CommonJS default export support
+module.exports = is;
+module.exports.default = is;
```
kevva: Doesn't `export default is` get compiled to this?

sindresorhus: No, it get's compiled to `exports.default = is`...

kevva: Ah, too bad.

ltetzlaff:
Cool, did not think of that either 👍

PR #31, opened 2017-11-08
Add is.boundFunction()

### LINKED ISSUE #30: Add `is.arrowFunction()`
Use-case: https://github.com/sindresorhus/is/issues/1#issuecomment-336866233

darrenscerri:
Should fix #30 . I think having `is.boundFunction()` is more appropriate than having `is.arrowFunction()` considering that the use-case is to check if a function is already bound or not.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -96,6 +96,21 @@ is.asyncFunction(() => {});
 // => false
 ```
 
+##### .boundFunction(value)
+
+Returns `true` for any `bound` function.
+
+```js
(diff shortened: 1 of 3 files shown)


brandon93s:
I agree that checking if a function is already bound fits the use case here better than `is.arrowFunction()`. If it's already bound, you know that thisArg is ignored with any invocation of `call` or `apply`. 👍

brandon93s:
Looks great! Thanks @darrenscerri.

PR #34, opened 2017-11-19
Add `is.nodeStream()`

### LINKED ISSUE #33: Add `is.nodeStream()`
I don't want to copy all the methods from [`is-stream`](https://github.com/sindresorhus/is-stream), but the main method is something I often need.

We can just copy this: https://github.com/sindresorhus/is-stream/blob/0342990d78787996ee0f98107c90497fe16593cc/index.js#L3-L6

brandon93s:
Closes #33


sindresorhus:
That was fast. Looks great :)

PR #38, opened 2017-12-09
Add `is.directInstanceOf()`

### LINKED ISSUE #36: `directInstanceOf()` method
Sometimes it's useful to be able to check if something is a direct instance of a class. `instanceOf` also matches any subclasses of that type, which is not always desirable.

```js
const directInstanceOf  = (instance, klass) => Object.getPrototypeOf(instance) === klass.prototype;
```

It's not a lot of code, but I always forget how to do it properly, so would be nice to have a check for this.

brandon93s:
- Add `is.directInstanceOf()`
- Using `klass` nomenclature pending [discussion](https://github.com/sindresorhus/is/issues/36#issuecomment-350423158)

-----

Closes #36

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -162,6 +162,8 @@ namespace is { // tslint:disable-line:no-namespace
 	export const arrayBuffer = isObjectOfType(TypeName.ArrayBuffer);
 	export const sharedArrayBuffer = isObjectOfType(TypeName.SharedArrayBuffer);
 
+	// TODO: Remove `object` checks when targeting ES2015 or higher
+	// See `Notes`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getPrototypeOf
 	export const directInstanceOf = (instance: any, klass: any) => object(instance) && object(klass) && Object.getPrototypeOf(instance) === klass.prototype;
 
 	export const truthy = (value: any) => Boolean(value);
--- a/source/index.ts
+++ b/source/index.ts
@@ -162,6 +162,8 @@ namespace is { // tslint:disable-line:no-namespace
 	export const arrayBuffer = isObjectOfType(TypeName.ArrayBuffer);
 	export const sharedArrayBuffer = isObjectOfType(TypeName.SharedArrayBuffer);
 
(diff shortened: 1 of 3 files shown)


source/index.ts
```diff
@@ -162,6 +162,8 @@ namespace is { // tslint:disable-line:no-namespace
 	export const arrayBuffer = isObjectOfType(TypeName.ArrayBuffer);
 	export const sharedArrayBuffer = isObjectOfType(TypeName.SharedArrayBuffer);
 
+	export const directInstanceOf = (instance: any, klass: any) => object(instance) && object(klass) && Object.getPrototypeOf(instance) === klass.prototype;
```
brandon93s: ES5 will throw a `TypeError` if `instance` isn't an object. ES2015 will coerced it to an object, and not throw an error. To support the lowest common denominator, I've required `instance` to be an `object` here to cover our bases and protect against an exception. I could be persuaded otherwise though if there is other thoughts on this.

source/tests/test.ts
```diff
@@ -507,6 +507,17 @@ test('is.dataView', t => {
...
 
+test('is.directInstanceOf', t => {
+	const error = new Error();
+	const errorSubclass = new ErrorSubclassFixture();
+
+	t.true(m.directInstanceOf(error, Error));
+	t.true(m.directInstanceOf(errorSubclass, ErrorSubclassFixture));
+
+	t.false(m.directInstanceOf(error, ErrorSubclassFixture));
+	t.false(m.directInstanceOf(errorSubclass, Error));
```
brandon93s: `t.false(m.directInstanceOf(errorSubclass, Error));` is really the use case here.

Included `t.false(m.directInstanceOf(error, ErrorSubclassFixture));` for completeness.

source/index.ts
brandon93s: [ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getPrototypeOf)

sindresorhus: Can you add a `TODO` comment about removing the object wrappers when we target only ES2015?

PR #43, opened 2018-03-29
Add type guards

SamVerschueren:
I don't have much time now, will have another look tomorrow if we can do some more simplifications with type guards.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -27,6 +27,32 @@ is.number(6);
 //=> true
 ```
 
+When using `is` together with TypeScript, [type guards](http://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) are being used to infer the correct type inside if-else statements.
+
+```ts
+import is from '@sindresorhus/is';
+
+const padLeft = (value: string, padding: string | number) => {
+	if (is.number(padding)) {
+		// `padding` is typed as `number`
+		return Array(padding + 1).join(' ') + value;
+	}
+
+	if (is.string(padding)) {
+		// `padding` is typed as `string`
+		return padding + value;
+	}
+
+	throw new TypeError(`Expected 'padding' to be of type 'string' or 'number', got '${is(padding)}'.`);
+}
+
+padLeft('🦄', 3);
+//=> '   🦄'
+
+padLeft('🦄', '🌈');
+//=> '🌈🦄'
+```
+
 
 ## API
 
--- a/source/index.ts
+++ b/source/index.ts
@@ -127,18 +127,18 @@ namespace is { // tslint:disable-line:no-namespace
 
 	export const nullOrUndefined = (value: any): value is null | undefined => null_(value) || undefined(value);
 	export const object = (value: any) => !nullOrUndefined(value) && (function_(value) || isObject(value));
-	export const iterable = <T = any>(value: any): value is Iterator<T> => !nullOrUndefined(value) && function_(value[Symbol.iterator]);
+	export const iterable = (value: any): value is Iterator<any> => !nullOrUndefined(value) && function_(value[Symbol.iterator]);
 	export const generator = (value: any): value is Generator => iterable(value) && function_(value.next) && function_(value.throw);
 
 	export const nativePromise = isObjectOfType<Promise<any>>(TypeName.Promise);
 
-	const hasPromiseAPI = <T = any>(value: any): value is Promise<T> =>
+	const hasPromiseAPI = (value: any): value is Promise<any> =>
 		!null_(value) &&
 		isObject(value) &&
 		function_(value.then) &&
 		function_(value.catch);
 
-	export const promise = <T = any>(value: any): value is Promise<T> => nativePromise(value) || hasPromiseAPI<T>(value);
+	export const promise = (value: any): value is Promise<any> => nativePromise(value) || hasPromiseAPI(value);
 
 	const isFunctionOfType = <T>(type: TypeName) => isObjectOfType<T>(type);
 	export const generatorFunction = isFunctionOfType<GeneratorFunction>(TypeName.GeneratorFunction);
--- a/source/index.ts
+++ b/source/index.ts
@@ -1,5 +1,12 @@
 import * as util from 'util';
 
+type TypedArray = Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array | Int32Array | Uint32Array | Float32Array | Float64Array;
+type Primitive = null | undefined | string | number | boolean | Symbol;
+
+export interface ArrayLike {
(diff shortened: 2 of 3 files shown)


SamVerschueren:
Added some more typeguards. The nice thing is that TypeScript now uses the correct type when using `is` inside an `if-else` statement for example.

```ts
if (is.nativePromise(x)) {
   // `x` is now typed as Promise
}
```

I believe I can do some more improvements, so PR is WIP.

source/index.ts
```diff
@@ -105,60 +112,60 @@ namespace is { // tslint:disable-line:no-namespace
...
 
 	export const array = Array.isArray;
 	export const buffer = Buffer.isBuffer;
 
-	export const nullOrUndefined = (value: any) => null_(value) || undefined(value);
+	export const nullOrUndefined = (value: any): value is null | undefined => null_(value) || undefined(value);
 	export const object = (value: any) => !nullOrUndefined(value) && (function_(value) || isObject(value));
-	export const iterable = (value: any) => !nullOrUndefined(value) && function_(value[Symbol.iterator]);
-	export const generator = (value: any) => iterable(value) && function_(value.next) && function_(value.throw);
+	export const iterable = <T = any>(value: any): value is Iterator<T> => !nullOrUndefined(value) && function_(value[Symbol.iterator]);
```
SamVerschueren: This allows you to write the following

```ts
if(is.iterable<string>(foo)) {
   const result = foo.next();
   // `result` is now an `IteratorResult` of type `string`
}
```

If you want I can remove the generic though as it's not really part of the check. It actually just infers the type of the items inside the Iterator.

source/index.ts
```diff
@@ -105,60 +112,60 @@ namespace is { // tslint:disable-line:no-namespace
...
+	export const generator = (value: any): value is Generator => iterable(value) && function_(value.next) && function_(value.throw);
 
-	export const nativePromise = isObjectOfType(TypeName.Promise);
+	export const nativePromise = isObjectOfType<Promise<any>>(TypeName.Promise);
```
SamVerschueren: I couldn't figure out how I could make this one generic, like the `is.promise` check below. Might be weird to be able to pass in a type to `.promise` but not to `.nativePromise`.

tsconfig.json
```diff
@@ -7,6 +7,11 @@
 		"declaration": true,
 		"pretty": true,
 		"newLine": "lf",
+		"lib": [
```
SamVerschueren: Had to set this to be able to use `SharedArrayBuffer`. Should we use `es2015` instead because that's our main target?

SamVerschueren:
Added some inline comments. I'm not sure about the generic stuff though. Feel free to provide feedback!

sindresorhus:
> Added some more typeguards. The nice thing is that TypeScript now uses the correct type when using is inside an if-else statement for example.

Could you document this? Seems like a nice feature users might not be aware of.

tsconfig.json
sindresorhus: > Should we use es2015 instead because that's our main target?

Yes

source/index.ts
```diff
@@ -1,5 +1,12 @@
 import * as util from 'util';
 
+type TypedArray = Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array | Int32Array | Uint32Array | Float32Array | Float64Array;
```
ysfjwd: if we do it like 

```
type TypedArray = Array<Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array | Int32Array | Uint32Array | Float32Array | Float64Array>;
```

then typescript will not raise 'cannot find signature on union types' for calling normal functions like `reduce`, `map`, `indexOf` etc.

ysfjwd: if you want, i can add a commit to this.

SamVerschueren: Although it fixes the issue, the type is incorrect. Now `TypedArray` is defined as an array of `Int8Array` or `Uint8Array`, etc.

What does work however, is the following

```ts
type TypedArray = (Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array | Int32Array | Uint32Array | Float32Array | Float64Array) & Array<number>;
```

This is ugly though. Will have a look as to why this error is caused.

ysfjwd: AFAIK, the error message "Cannot invoke an expression whose type lacks a call signature"" is by design. see microsoft/typescript#7294 .

SamVerschueren:
I thought more about the generic functions we expose, e.g. the following

```ts
if (is.promise<number>(foo)) {
   const value = await foo;
   // value is of type number
}
```

I don't think we should do that. The only thing that `is` does, is check the type and infer the type correctly. It's not up to `is` to do more than that. If users type there values correctly, all of it will go automatically.

```ts
const foo: Promise<number> | undefined = someValue;

if (is.promise(foo)) {
   const value = await foo;
   // value is of type number
}
```

If they didn't specify the types correctly, they can always type it themselves.

```ts
// foo is of type any

if (is.promise(foo)) {
   const value: number = await foo;
   // value is of type number
}
```

So I believe we should get rid of the generics to the outside world because it's not up to `is` to enforce the type the wrapper resolves to. We currently only have generics exposed for `is.promise` and `is.iterable`.

ysfjwd:
@SamVerschueren thank you for kind reply. It makes sense. I saw the call to review by @sindresorhus on twitter, and looked at it. this particular gave me quite a headache a few weeks ago while upgrading a projecting from `js` to `ts`. thank you again.

source/index.ts
SamVerschueren: I think I'm going to leave it as is for now. It's explained in the issue that you referenced, this is by design. If you know what type it will be, it would be safer to use `is.int8Array` or one of the others. Thanks for letting me know, learned something new today :).

readme.md
```diff
@@ -27,6 +27,32 @@ is.number(6);
 //=> true
 ```
 
+When using `is` together with TypeScript, [type guards](http://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) are being used to infer the correct type inside if-else statements.
```
SamVerschueren: In VSCode, type inference also works in JavaScript. Don't think we should document this as it's IDE specific.

SamVerschueren:
@sindresorhus Processed feedback. Let me know if you have some more feedback.

sindresorhus:
> I don't think we should do that. The only thing that is does, is check the type and infer the type correctly. It's not up to is to do more than that. If users type there values correctly, all of it will go automatically.

Late answer, but yes, I agree.

sindresorhus:
Yay! Such strong typing 💪 Quality work as always, Sam.

PR #49, opened 2018-04-29
 Improve TS code by adding generics and removing `any` where applicable

ltetzlaff:
This is in response to: https://github.com/sindresorhus/ow/pull/68#issuecomment-385262882

The switch at the top is debatable but should be faster and is also shorter 🙆
Generally it is more an improvement for TS-users and we directly see if our typings are not set correctly, as to be seen in the tests for instance 🕵️
(If you're also interested why `() => {}` is ArrayLike (=== has a length) check this out: [Function.length](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length))

Please bear with me for throwing plenty of those `any`s in in the first place: [Blame](https://github.com/sindresorhus/is/blame/master/source/index.ts) 🙈🙉

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -111,7 +111,7 @@ function is(value: any): TypeName { // tslint:disable-line:only-arrow-functions
 }
 
 namespace is { // tslint:disable-line:no-namespace
-	const isObject = (value: object) => typeof value === 'object'; // tslint:disable-line:strict-type-predicates
+	const isObject = (value: any): value is object => typeof value === 'object';
 
 	// tslint:disable:variable-name
 	export const undefined = isOfType<undefined>('undefined');
@@ -137,7 +137,7 @@ namespace is { // tslint:disable-line:no-namespace
 
 	const hasPromiseAPI = (value: any): value is Promise<any> =>
 		!null_(value) &&
-		isObject(value) &&
+		isObject(value) as any &&
 		function_(value.then) &&
 		function_(value.catch);
 
@@ -246,7 +246,7 @@ namespace is { // tslint:disable-line:no-namespace
 	export const domElement = (value: any): value is DomElement => object(value) as any && value.nodeType === NODE_TYPE_ELEMENT && string(value.nodeName) &&
 		!plainObject(value) && DOM_PROPERTIES_TO_CHECK.every(property => property in value);
 
-	export const nodeStream = (value: any): value is NodeStream => !nullOrUndefined(value) && isObject(value) && function_(value.pipe);
+	export const nodeStream = (value: any): value is NodeStream => !nullOrUndefined(value) && isObject(value) as any && function_(value.pipe);
 
 	export const infinite = (value: any) => value === Infinity || value === -Infinity;
 
--- a/source/index.ts
+++ b/source/index.ts
@@ -7,10 +7,12 @@ export interface ArrayLike {
 	length: number;
 }
 
-type Class<T = any> = { new(...args: any[]): T; };
+export interface Class<T = any> {
+	new(...args: any[]): T;
+}
 
-type DomElement = object & { nodeType: 1; nodeName: string }
-type NodeStream = object & { pipe: Function }
+type DomElement = object & { nodeType: 1; nodeName: string };
+type NodeStream = object & { pipe: Function };
 
 export const enum TypeName {
 	null = 'null',
@@ -109,7 +111,7 @@ function is(value: any): TypeName { // tslint:disable-line:only-arrow-functions
 }
 
 namespace is { // tslint:disable-line:no-namespace
-	const isObject = (value: object) => typeof value === 'object';
+	const isObject = (value: object) => typeof value === 'object'; // tslint:disable-line:strict-type-predicates
 
 	// tslint:disable:variable-name
 	export const undefined = isOfType<undefined>('undefined');
@@ -241,8 +243,8 @@ namespace is { // tslint:disable-line:no-namespace
 		'nodeValue'
 	];
 
-	export const domElement = (value: any): value is DomElement => value.nodeType === NODE_TYPE_ELEMENT && string(value.nodeName) &&
-		!plainObject(value) && DOM_PROPERTIES_TO_CHECK.every(property => property in value) && object(value);
+	export const domElement = (value: any): value is DomElement => object(value) as any && value.nodeType === NODE_TYPE_ELEMENT && string(value.nodeName) &&
+		!plainObject(value) && DOM_PROPERTIES_TO_CHECK.every(property => property in value);
 
 	export const nodeStream = (value: any): value is NodeStream => !nullOrUndefined(value) && isObject(value) && function_(value.pipe);
 
--- a/source/index.ts
+++ b/source/index.ts
@@ -182,7 +182,7 @@ namespace is { // tslint:disable-line:no-namespace
 	export const integer = (value: number): value is number => Number.isInteger(value);
 	export const safeInteger = (value: number): value is number => Number.isSafeInteger(value);
 
-	export const plainObject = (value: any) => {
+	export const plainObject = (value: object) => {
 		// From: https://github.com/sindresorhus/is-plain-obj/blob/master/index.js
 		let prototype;
 
@@ -239,7 +239,7 @@ namespace is { // tslint:disable-line:no-namespace
 	export const domElement = (value: any) => object(value) && value.nodeType === NODE_TYPE_ELEMENT && string(value.nodeName) &&
 		!plainObject(value) && DOM_PROPERTIES_TO_CHECK.every(property => property in value);
 
-	export const nodeStream = (value: any) => !nullOrUndefined(value) && isObject(value) && function_(value.pipe);
+	export const nodeStream = (value: object) => !nullOrUndefined(value) && isObject(value) && function_(value.pipe);
 
 	export const infinite = (value: number) => value === Infinity || value === -Infinity;
 
--- a/source/index.ts
+++ b/source/index.ts
@@ -129,7 +129,7 @@ namespace is { // tslint:disable-line:no-namespace
 
 	export const nullOrUndefined = (value: any): value is null | undefined => null_(value) || undefined(value);
 	export const object = (value: any): value is object => !nullOrUndefined(value) && (function_(value) || isObject(value));
-	export const iterable = <T = any>(value: any): value is IterableIterator<T> => !nullOrUndefined(value) && function_(value[Symbol.iterator]);
+	export const iterable = (value: any): value is IterableIterator<any> => !nullOrUndefined(value) && function_(value[Symbol.iterator]);
 	export const generator = (value: any): value is Generator => iterable(value) && function_(value.next) && function_(value.throw);
 
 	export const nativePromise = (value: any): value is Promise<any> =>
@@ -150,10 +150,10 @@ namespace is { // tslint:disable-line:no-namespace
 	export const regExp = isObjectOfType<RegExp>(TypeName.RegExp);
 	export const date = isObjectOfType<Date>(TypeName.Date);
 	export const error = isObjectOfType<Error>(TypeName.Error);
-	export const map = (value: Map<any, any>): value is Map<any, any> => isObjectOfType<Map<any, any>>(TypeName.Map)(value);
-	export const set = (value: Set<any>): value is Set<any> => isObjectOfType<Set<any>>(TypeName.Set)(value);
-	export const weakMap = (value: WeakMap<any, any>): value is WeakMap<any, any> => isObjectOfType<WeakMap<any, any>>(TypeName.WeakMap)(value);
-	export const weakSet = (value: WeakSet<any>): value is WeakSet<any> => isObjectOfType<WeakSet<any>>(TypeName.WeakSet)(value);
+	export const map = (value: any): value is Map<any, any> => isObjectOfType<Map<any, any>>(TypeName.Map)(value);
+	export const set = (value: any): value is Set<any> => isObjectOfType<Set<any>>(TypeName.Set)(value);
+	export const weakMap = (value: any): value is WeakMap<any, any> => isObjectOfType<WeakMap<any, any>>(TypeName.WeakMap)(value);
+	export const weakSet = (value: any): value is WeakSet<any> => isObjectOfType<WeakSet<any>>(TypeName.WeakSet)(value);
 
 	export const int8Array = isObjectOfType<Int8Array>(TypeName.Int8Array);
 	export const uint8Array = isObjectOfType<Uint8Array>(TypeName.Uint8Array);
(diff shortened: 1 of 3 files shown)


sindresorhus:
What was the original argument for using `any` in all of the methods? I remember we had a discussion about it and ended up with that. @SamVerschueren Do you remember/

source/index.ts
```diff
@@ -118,40 +113,40 @@ namespace is { // tslint:disable-line:no-namespace
 	export const function_ = isOfType<Function>('function');
 	export const null_ = (value: any): value is null => value === null;
```
sindresorhus: `any` => `null`?

source/index.ts
```diff
@@ -246,22 +241,22 @@ namespace is { // tslint:disable-line:no-namespace
 
 	export const nodeStream = (value: any) => !nullOrUndefined(value) && isObject(value) && function_(value.pipe);
```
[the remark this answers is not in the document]
ltetzlaff: `plainObject` and `nodeStream` definitely!
`domElement` is a little tricky since it would be `object & { nodeType: 1, nodeName: string, etc.. }`

ltetzlaff:
Basically you can still do that if you're a JS user, for TS people that won't really help because the compiler will tell them if what they're trying to do is not allowed and will most likely be used for people to check `CLI inputs`, `JSON.parse()` etc. 

For those cases I'd say working with `const a = JSON.parse("whatever")` which is typed as `any` still allows throwing them into our methods to further decide what we have there.

source/index.ts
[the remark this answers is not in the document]
sindresorhus: > domElement is a little tricky since it would be object & { nodeType: 1, nodeName: string, etc.. }

But even just `object` (without the `nodeType` etc) for `domElement` would be better than `any`, right?

source/index.ts
```diff
@@ -109,49 +104,49 @@ function is(value: any): TypeName { // tslint:disable-line:only-arrow-functions
...
+	const isObject = (value: object) => typeof value === 'object'; // tslint:disable-line:strict-type-predicates
 
 	// tslint:disable:variable-name
 	export const undefined = isOfType<undefined>('undefined');
 	export const string = isOfType<string>('string');
 	export const number = isOfType<number>('number');
 	export const function_ = isOfType<Function>('function');
-	export const null_ = (value: any): value is null => value === null;
+	export const null_ = (value: null | any): value is null => value === null;
 	export const class_ = (value: any) => function_(value) && value.toString().startsWith('class ');
```
sindresorhus: Could this one have a `class` type?

See:
- https://stackoverflow.com/questions/39392853/is-there-a-type-for-class-in-typescript-and-does-any-include-it
- https://stackoverflow.com/questions/39976329/typescript-type-to-represent-any-class

source/index.ts
```diff
@@ -109,49 +104,49 @@ function is(value: any): TypeName { // tslint:disable-line:only-arrow-functions
...
 	export const buffer = Buffer.isBuffer;
 
-	export const nullOrUndefined = (value: any): value is null | undefined => null_(value) || undefined(value);
-	export const object = (value: any) => !nullOrUndefined(value) && (function_(value) || isObject(value));
-	export const iterable = (value: any): value is Iterator<any> => !nullOrUndefined(value) && function_(value[Symbol.iterator]);
-	export const generator = (value: any): value is Generator => iterable(value) && function_(value.next) && function_(value.throw);
+	export const nullOrUndefined = (value: null | undefined | any): value is null | undefined => null_(value) || undefined(value);
+	export const object = (value: object) => !nullOrUndefined(value) && (function_(value) || isObject(value));
+	export const iterable = <T = any>(value: IterableIterator<T>): value is IterableIterator<T> => !nullOrUndefined(value) && function_(value[Symbol.iterator]);
+	export const generator = (value: Generator): value is Generator => iterable(value as IterableIterator<any>) && function_(value.next) && function_(value.throw);
```
sindresorhus: What's the point of using `value is Generator` when we already define `value: Generator` as a parameter?

source/index.ts
```diff
@@ -172,7 +167,7 @@ namespace is { // tslint:disable-line:no-namespace
 	export const truthy = (value: any) => Boolean(value);
```
sindresorhus: Same as https://github.com/sindresorhus/is/pull/49/files?w=1#r185387543 for `directInstanceOf`'s `klass` argument. (I can't comment on the exact line.

source/index.ts
```diff
@@ -241,27 +236,27 @@ namespace is { // tslint:disable-line:no-namespace
...
 
 	const isAbsoluteMod2 = (value: number) => (rem: number) => integer(rem) && Math.abs(rem % 2) === value;
 	export const even = isAbsoluteMod2(0);
 	export const odd = isAbsoluteMod2(1);
 
-	const isWhiteSpaceString = (value: any) => string(value) && /\S/.test(value) === false;
-	const isEmptyStringOrArray = (value: any) => (string(value) || array(value)) && value.length === 0;
+	const isWhiteSpaceString = (value: string) => string(value) && /\S/.test(value) === false;
+	const isEmptyStringOrArray = (value: string | any[]) => (string(value) || array(value)) && value.length === 0;
 	const isEmptyObject = (value: any) => !map(value) && !set(value) && object(value) && Object.keys(value).length === 0;
```
sindresorhus: `any` => `object`?

source/index.ts
```diff
@@ -241,27 +236,27 @@ namespace is { // tslint:disable-line:no-namespace
...
 	const isEmptyMapOrSet = (value: any) => (map(value) || set(value)) && value.size === 0;
```
sindresorhus: `any` => `Map<any, any> | Set<any>`?

SamVerschueren:
> What was the original argument for using any in all of the methods? 

Not sure anymore, don't think there was a specific reason.

source/index.ts
```diff
@@ -109,49 +104,49 @@ function is(value: any): TypeName { // tslint:disable-line:only-arrow-functions
...
-	const isObject = (value: any) => typeof value === 'object';
+	const isObject = (value: object) => typeof value === 'object'; // tslint:disable-line:strict-type-predicates
 
 	// tslint:disable:variable-name
 	export const undefined = isOfType<undefined>('undefined');
 	export const string = isOfType<string>('string');
 	export const number = isOfType<number>('number');
 	export const function_ = isOfType<Function>('function');
-	export const null_ = (value: any): value is null => value === null;
+	export const null_ = (value: null | any): value is null => value === null;
```
SamVerschueren: Why `any`?

source/index.ts
```diff
@@ -109,49 +104,49 @@ function is(value: any): TypeName { // tslint:disable-line:only-arrow-functions
 }
 
 namespace is { // tslint:disable-line:no-namespace
-	const isObject = (value: any) => typeof value === 'object';
+	const isObject = (value: object) => typeof value === 'object'; // tslint:disable-line:strict-type-predicates
```
SamVerschueren: Why disabling `strict-type-predicates`? Tested it locally and doesn't seem to fail without disabling it.

source/index.ts
```diff
@@ -109,49 +104,49 @@ function is(value: any): TypeName { // tslint:disable-line:only-arrow-functions
...
 	// tslint:enable:variable-name
 
 	export const array = Array.isArray;
 	export const buffer = Buffer.isBuffer;
 
-	export const nullOrUndefined = (value: any): value is null | undefined => null_(value) || undefined(value);
-	export const object = (value: any) => !nullOrUndefined(value) && (function_(value) || isObject(value));
-	export const iterable = (value: any): value is Iterator<any> => !nullOrUndefined(value) && function_(value[Symbol.iterator]);
-	export const generator = (value: any): value is Generator => iterable(value) && function_(value.next) && function_(value.throw);
+	export const nullOrUndefined = (value: null | undefined | any): value is null | undefined => null_(value) || undefined(value);
```
SamVerschueren: Same as above, why `any`?

source/index.ts
```diff
@@ -109,49 +104,49 @@ function is(value: any): TypeName { // tslint:disable-line:only-arrow-functions
...
+	export const object = (value: object) => !nullOrUndefined(value) && (function_(value) || isObject(value));
+	export const iterable = <T = any>(value: IterableIterator<T>): value is IterableIterator<T> => !nullOrUndefined(value) && function_(value[Symbol.iterator]);
```
SamVerschueren: We discussed this and we decided not to do this. I explain [here](https://github.com/sindresorhus/is/pull/43#issuecomment-377780830) why.

> I don't think we should do that. The only thing that is does, is check the type and infer the type correctly. It's not up to is to do more than that. If users type there values correctly, all of it will go automatically.

source/index.ts
SamVerschueren: That might be one of the reasons I went for `(value: any)`. If you already know the type, what's the point in using `is`. It could lead to weird implementations

```ts
const x = some3rdPartyCall();

// Cast it because otherwise we can't test it
if (is.generator(x as Generator)) {
   // Now we know at runtime that `x` is a generator
}
```

source/index.ts
```diff
@@ -109,49 +104,49 @@ function is(value: any): TypeName { // tslint:disable-line:only-arrow-functions
...
+	export const generator = (value: Generator): value is Generator => iterable(value as IterableIterator<any>) && function_(value.next) && function_(value.throw);
 
-	export const nativePromise = isObjectOfType<Promise<any>>(TypeName.Promise);
+	export const nativePromise = <T = any>(value: Promise<T>): value is Promise<T> =>
```
SamVerschueren: Same as above, remove the generic.

source/index.ts
```diff
@@ -109,49 +104,49 @@ function is(value: any): TypeName { // tslint:disable-line:only-arrow-functions
...
+		isObjectOfType<Promise<T>>(TypeName.Promise)(value);
 
-	const hasPromiseAPI = (value: any): value is Promise<any> =>
+	const hasPromiseAPI = <T = any>(value: Promise<T>): value is Promise<T> =>
```
SamVerschueren: Same as above, remove the generic.

source/index.ts
```diff
@@ -109,49 +104,49 @@ function is(value: any): TypeName { // tslint:disable-line:only-arrow-functions
...
 		!null_(value) &&
 		isObject(value) &&
 		function_(value.then) &&
 		function_(value.catch);
 
-	export const promise = (value: any): value is Promise<any> => nativePromise(value) || hasPromiseAPI(value);
+	export const promise = <T = any>(value: Promise<T>): value is Promise<T> => nativePromise<T>(value) || hasPromiseAPI<T>(value);
```
SamVerschueren: Same as above, remove the generic.

source/index.ts
```diff
@@ -109,49 +104,49 @@ function is(value: any): TypeName { // tslint:disable-line:only-arrow-functions
...
 	export const date = isObjectOfType<Date>(TypeName.Date);
 	export const error = isObjectOfType<Error>(TypeName.Error);
-	export const map = isObjectOfType<Map<any, any>>(TypeName.Map);
-	export const set = isObjectOfType<Set<any>>(TypeName.Set);
-	export const weakMap = isObjectOfType<WeakMap<any, any>>(TypeName.WeakMap);
-	export const weakSet = isObjectOfType<WeakSet<any>>(TypeName.WeakSet);
+	export const map = <T1 = any, T2 = any>(value: Map<T1, T2>): value is Map<T1, T2> => isObjectOfType<Map<T1, T2>>(TypeName.Map)(value);
+	export const set = <T = any>(value: Set<T>): value is Set<T> => isObjectOfType<Set<T>>(TypeName.Set)(value);
+	export const weakMap = <T1 extends object = any, T2 = any>(value: WeakMap<T1, T2>): value is WeakMap<T1, T2> => isObjectOfType<WeakMap<T1, T2>>(TypeName.WeakMap)(value);
+	export const weakSet = <T extends object = any>(value: WeakSet<T>): value is WeakSet<T> => isObjectOfType<WeakSet<T>>(TypeName.WeakSet)(value);
```
SamVerschueren: All of these `map`, `set`, `weakMap`, `weakSet` should not use generics. See above for details.

source/index.ts
```diff
@@ -241,27 +236,27 @@ namespace is { // tslint:disable-line:no-namespace
...
+	const isEmptyStringOrArray = (value: string | any[]) => (string(value) || array(value)) && value.length === 0;
 	const isEmptyObject = (value: any) => !map(value) && !set(value) && object(value) && Object.keys(value).length === 0;
 	const isEmptyMapOrSet = (value: any) => (map(value) || set(value)) && value.size === 0;
 
 	export const empty = (value: any) => falsy(value) || isEmptyStringOrArray(value) || isEmptyObject(value) || isEmptyMapOrSet(value);
 	export const emptyOrWhitespace = (value: any) => empty(value) || isWhiteSpaceString(value);
 
-	type ArrayMethod = (fn: (value: any, index: number, array: any[]) => boolean, thisArg?: any) => boolean;
-	const predicateOnArray = (method: ArrayMethod, predicate: any, values: any[]) => {
+	type ArrayMethod = <T>(fn: (value: T, index: number, array: T[]) => boolean, thisArg?: any) => boolean;
```
SamVerschueren: Same as above, remove the generic.

source/tests/test.ts
```diff
@@ -582,15 +582,15 @@ test('is.plainObject', t => {
...
-	t.false(m.iterable({}));
+	t.true(m.iterable('' as any));
+	t.true(m.iterable([] as any));
+	t.true(m.iterable(new Map() as any));
+	t.false(m.iterable(null as any));
+	t.false(m.iterable(undefined as any));
+	t.false(m.iterable(0 as any));
+	t.false(m.iterable(NaN as any));
+	t.false(m.iterable(Infinity as any));
+	t.false(m.iterable({} as any));
```
SamVerschueren: As explained above, this is the reason why I'm not sure if we should strongly type the `value` arguments. My gut feeling says `I just want to throw values to the library, and it should either return true or false`. Always having to typecast it to the specific type either assumes you already know what it is, or otherwise you need type it as `any`.

Another example would be a function which either returns a `Promise` or an `Observable`.

```ts
const value = someFunction();

// `value` is of type `Promise` | `Observable`, but have to cast it explicitely to `Promise` or any.
if (is.promise(value as Promise)) {
   // ...
}
```

ltetzlaff:
According to @SamVerschueren's review I refactored to use `(value: any)` consistently, only `inRange` and `isAbsoluteMod2` are still typed to expect numbers because we do not check every parameter for `number(v)` explicitly, ultimately based on these two very valid arguments:

https://github.com/sindresorhus/is/pull/49#discussion_r185393242
https://github.com/sindresorhus/is/pull/49#discussion_r185392646

To improve the type safety of the result I added `(): value is T` where missing, though.

sindresorhus:
Thanks again for such great work @ltetzlaff :)

PR #50, opened 2018-05-02
Add `is.observable`

SamVerschueren:
Added `is.observable` which uses the exact same implementation as [is-observable](https://github.com/sindresorhus/is-observable).

@sindresorhus Would be fine to use `is-observable` directly as a dependency though.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,6 +5,7 @@
 		"module": "commonjs",
 		"moduleResolution": "node",
 		"declaration": true,
+		"skipLibCheck": true,
 		"pretty": true,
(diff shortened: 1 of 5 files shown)


tsconfig.json
```diff
@@ -5,6 +5,7 @@
 		"module": "commonjs",
 		"moduleResolution": "node",
 		"declaration": true,
+		"skipLibCheck": true,
```
SamVerschueren: If we don't do this, the following error appears when compiling.

screen shot 2018-05-02 at 22 16 44

PR #59, opened 2018-07-07
Add `is.asyncIterable`

### LINKED ISSUE #56: `is.asyncIterable()` method
Something like:

```js
typeof Symbol.asyncIterator !== 'undefined' && typeof value[Symbol.asyncIterator] === 'function'
```

arturovt:
Added `is.asyncIterable`

Fixes #56

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -48,8 +48,8 @@ export const enum TypeName {
 	ArrayBuffer = 'ArrayBuffer',
 	SharedArrayBuffer = 'SharedArrayBuffer',
 	DataView = 'DataView',
-    Promise = 'Promise',
-    URL = 'URL'
+	Promise = 'Promise',
+	URL = 'URL'
 }
 
 const toString = Object.prototype.toString;
--- a/source/index.ts
+++ b/source/index.ts
@@ -48,8 +48,7 @@ export const enum TypeName {
 	ArrayBuffer = 'ArrayBuffer',
 	SharedArrayBuffer = 'SharedArrayBuffer',
 	DataView = 'DataView',
-	Promise = 'Promise',
-	URL = 'URL'
+	Promise = 'Promise'
 }
 
 const toString = Object.prototype.toString;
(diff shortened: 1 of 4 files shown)


source/index.ts
```diff
@@ -136,6 +136,7 @@ namespace is { // tslint:disable-line:no-namespace
 	export const nullOrUndefined = (value: any): value is null | undefined => null_(value) || undefined(value);
 	export const object = (value: any): value is object => !nullOrUndefined(value) && (function_(value) || isObject(value));
 	export const iterable = (value: any): value is IterableIterator<any> => !nullOrUndefined(value) && function_(value[Symbol.iterator]);
+	export const asyncIterable = (value: any): value is AsyncIterableIterator<any> => !nullOrUndefined(Symbol.asyncIterator) && function_(value[Symbol.asyncIterator]);
```
sindresorhus: You also need the `!nullOrUndefined(value)` check, like in the above `iterable` predicate.

It would currently fail on `t.false(m.asyncIterable(undefined));`

source/tests/test.ts
```diff
@@ -454,6 +455,14 @@ if (isNode8orHigher) {
 	});*/
 }
 
+if (isNode10orHigher) {
+	test('is.asyncIterable', t => {
+		t.true(m.asyncIterable(Object.create({
+			[Symbol.asyncIterator]: () => {} // tslint:disable-line:no-empty
+		})));
+	});
+}
```
sindresorhus: Move this below the `is.iterable` test to keep the order. This also needs more assertions, use the same ones as used in the `is.iterable` test.

source/tests/test.ts
```diff
@@ -454,6 +455,14 @@ if (isNode8orHigher) {
 	});*/
 }
 
+if (isNode10orHigher) {
+	test('is.asyncIterable', t => {
+		t.true(m.asyncIterable(Object.create({
```
sindresorhus: I don't see the point of using `Object.create()` here. It could just be a plain object literal.

arturovt: @sindresorhus, Yep, you're right

sindresorhus:
It also needs to be added to the readme.

sindresorhus:
Looks good. Nice work, @arturovt :)

PR #60, opened 2018-07-08
Add `is.urlInstance`

### LINKED ISSUE #58: Detect `new URL()` instance
`is.url()` might be too confusing though. Maybe `is.urlObject()`? Feedback welcome.

arturovt:
Fixes #58

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -48,8 +48,8 @@ export const enum TypeName {
 	ArrayBuffer = 'ArrayBuffer',
 	SharedArrayBuffer = 'SharedArrayBuffer',
 	DataView = 'DataView',
-    Promise = 'Promise',
-    URL = 'URL'
+	Promise = 'Promise',
+	URL = 'URL'
 }
 
 const toString = Object.prototype.toString;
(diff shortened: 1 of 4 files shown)


sindresorhus:
Don't include commits from a different PR.

sindresorhus:
It needs to be added to the readme.

source/index.ts
```diff
@@ -176,6 +178,7 @@ namespace is { // tslint:disable-line:no-namespace
 	export const dataView = isObjectOfType<DataView>(TypeName.DataView);
 
 	export const directInstanceOf = <T>(instance: any, klass: Class<T>): instance is T => Object.getPrototypeOf(instance) === klass.prototype;
+	export const urlInstance = (value: any): value is URL => !nullOrUndefined(value) && directInstanceOf<URL>(value, URL);
```
sindresorhus: Using instance checking will not work across realms (frames in the browser or `vm` module in Node.js). 

A better solution would be to add `URL` as a type name here: https://github.com/arturovt/is/blob/7e71742a0f6aac5f44be0f2e00ed1619c75fd557/source/index.ts#L53 and then use `isObjectOfType`, like, for example, the `is.generatorFunction` check.

sindresorhus:
Also please include which issue this PR fixes in the PR body in the format of `Fixes #123` (This should always be done on PRs on GitHub).

source/index.ts
arturovt: @sindresorhus Agree

source/tests/test.ts
```diff
@@ -533,6 +534,16 @@ test('is.directInstanceOf', t => {
 	t.false(m.directInstanceOf(errorSubclass, Error));
 });
 
+test('is.urlInstance', t => {
+	const url = new URL('https://google.com');
```
brandon93s: http://www.example.com/ is maintained for [documentation](https://www.iana.org/domains/reserved) purposes. I recommend using them in examples, tests and documentation. 

Same for the readme.

Issue #53, opened 2018-05-25
is.lengthyArray


rarkins:
Would you support a PR for is.lengthyArray? Ie returns true if is.array(input) && input.length

sindresorhus:
This already exists as `!is.empty(array)`.

rarkins:
Thanks, that looks good

rarkins:
Actually, I don't think it's the same. For example, `!is.empty('b')` returns `true` while it is not an array.

brandon93s:
Using `is.array(x) && !is.empty(x)` reads nicely and I believe accomplishes the goal. If you're using it often, a local helper could support the use case:

```js
const isNonEmptyArray = x => is.array(x) && !is.empty(x) 
```

rarkins:
I was hoping that `is` would remove the need for commonly-used local helper functions and checking that `x` is a non-empty array is a fairly common use case. 

If the criteria for rejecting proposals is that they're composed of two other ones, then `.nullOrUndefined(value)` reads even nicer:

```js
const isNullOrUndefined => x => is.null(x) || is.undefined(x);
```

Another example (I think):
```js
const isEmptyObject => x => is.object(x) && is.empty(x);
```

sindresorhus:
I'm starting to think `is.empty()` was a mistake. I can't really see its use-case without first checking the type.

I'm leaning towards creating individual `is.emptyArray`/`is.nonEmptyArray` like methods for String, Array, Object, Map, Set.

Thoughts?

/// @SamVerschueren

SamVerschueren:
I agree that `is.empty()` doesn't really say what it checks. With `Ow`, we have the benefit that you have to specify the type first, so `ow.string.empty()` really makes sense. But with `is`, we don't have that so I think we should make it `ow.emptyString()`, `ow.emptyArray()`, ...

sindresorhus:
Alright. PR welcome to remove `is.empty` and add the mentioned methods.

ArfatSalman:
I am taking this up.

PR #61, opened 2018-07-11
Add is.emptyArray/is.nonEmptyArray and for other types #53

ArfatSalman:


sindresorhus:
Missing `nonEmpty` versions of the other types like `Object`, `Set`, etc.

ArfatSalman:
Ok. On it.

readme.md
```diff
@@ -295,13 +295,64 @@ Returns `true` if `value` is an even integer.
...
+##### .emptyObject(value)
+
+Returns `true` if the value is an object and `Object.keys(value)` length is 0. 
+
+Please note that `Object.keys` returns only enumerable properties. Hence something like this can happen - 
+
+```js
+const object1 = {};
+
+Object.defineProperty(a, 'property1', {
```
gioragutt: What's the purpose of the defineProperty in this example

readme.md
```diff
@@ -295,13 +295,64 @@ Returns `true` if `value` is an even integer.
...
+
+##### .emptyMap(value)
+
+Returns `true` if the value is an empty Map (`size` is 0). `false` if it is not zero.
+
+##### .nonEmptyMap(value)
+
+Returns `true` if the value is a Map and it's `size` is not 0. `false` if is 0.
+
+##### .emptyStringOrWhitespace(value)
```
gioragutt: Move this up, after the empty and non empty string methods

source/index.ts
```diff
@@ -265,12 +265,23 @@ namespace is { // tslint:disable-line:no-namespace
...
+	export const emptyObject = (value: any) => !map(value) && !set(value) && object(value) && Object.keys(value).length === 0;
+	export const nonEmptyObject = (value: any) => !map(value) && !set(value) && object(value) && Object.keys(value).length !== 0;
+
+	export const emptySet = (value: any) => set(value) && value.size === 0;
+	export const nonEmptySet = (value: any) => set(value) && value.size !== 0;
+
+	export const emptyMap = (value: any) => map(value) && value.size === 0;
+	export const nonEmptyMap = (value: any) => map(value) && value.size !== 0;
+
+	export const emptyStringOrWhitespace = (value: any) => emptyString(value) || isWhiteSpaceString(value);
```
gioragutt: Like in the docs, move next to the other string methods

source/tests/test.ts
```diff
@@ -739,40 +739,73 @@ test('is.odd', t => {
 	}
 });
 
-test('is.empty', t => {
-	t.true(m.empty(null));
-	t.true(m.empty(undefined));
+test('is.emptyArray', t => {
+	t.true(m.emptyArray([]));
```
gioragutt: Why not use testType for all these examples? The whole point of the change is to make sure that the methods work on their respective type only. This doesn't do that. Refers to all the empty/nonEmpty tests

readme.md
ArfatSalman: It is because `isEmptyObject` checks for only `enumerable` properties. Hence,  I added an example to show that an object can have non-enumerable properties and still the result will be empty.

readme.md
ArfatSalman: Ok.

source/index.ts
ArfatSalman: Ok.

readme.md
gioragutt: So change it so that defineProperty will run on `object1` and not on `a`

sindresorhus:
In the readme, instead of putting these under `#### Miscellaneous`, I think it would make sense to give them a separate sub-section called `#### Emptiness`.

sindresorhus:
es

source/tests/test.ts
ArfatSalman: @gioragutt

gioragutt: @ArfatSalman

You can see that in the `is.plainObject` tests:
```typescript
test('is.plainObject', t => {
	testType(t, 'plainObject', ['object', 'promise']);
});
```

The `object` and `promise` tests are excluded.
What `testType` does, it run the the test you specified in the [`types`](https://github.com/sindresorhus/is/blob/master/source/tests/test.ts#L28) array, and check that the given predicate works for the fixtures your provided, but not for all the others.

In `plainObject`'s case, there are tests for different methods that succeed (`object` and `promise`), thus they are excluded.

You should add `is.<yourMethod>` to the `types` array.
I would reconsider doing it to `nonEmpty` methods, since they might have a little too much exceptions.
Thus, for the `emptyX` tests, use the `testType` method and add your cases to the `types` array,
and for the `nonEmpty` you should write specific assertions.

@sindresorhus sounds good?

sindresorhus: @gioragutt 👍

ArfatSalman:
Hi @gioragutt 

I did exactly that. I created my own fixture in the types array, like this - 
```js
	['emptyString', {
		is: m.emptyString,
		fixtures: [
			'',
			String()
		]
	}]
```

and I tested it like this - 

```js
test('is.emptyString', t => {
	testType(t, 'emptyString');
});
```
However, it failed the test. Upon inspection, I found that the `string` type has a fixture `''`. `testType` expects all other fixtures to be `false` and the key's fixture to be `true`. That is, only the fixtures of `emptyString`  with `m.emptyString` should return true. However, the empty string in `string` fixture also returns true when testType was expecting it to be false. Which leads to the entire test case failing. 

Hence, I am not able to add the `emptyString` to the types array along with my fixtures. 

I maybe confused as to `testType` does, but this is what I gleamed from my inspection of testType.

gioragutt:
@ArfatSalman see testType's second parameter.

sindresorhus:
@ArfatSalman Still interested in finishing this?

ArfatSalman:
Yes. On it. Sorry for the delay.

ArfatSalman:
@gioragutt Please review the PR and let me know if there is something else that I can do. 
Thanks.

PR #66, opened 2018-09-26
Use is-buffer ponyfill for better browser support

SamVerschueren:
This is using an `isBuffer` ponyfill to make sure it works out of the box in browsers without the need to to `global.Buffer = require('buffer');`


gioragutt:
Nice, looks good. @sindresorhus can be merged, no?

PR #67, opened 2018-10-24
Add `is.numericString()`

### LINKED ISSUE #54: `is.numericString()`
Sometimes I need to check if a string is a valid number. I guess we could just do `!Number.isNaN(Number(string));`

itaisteinherz:
I implemented the method based on the discussion over at #54 (and this closes #54).
Note: I added the check `!nullOrUndefined(value)`, since without it `is.numericString(null)` would return `true` (because `Number(null)` is `0`).

This is the first time I ever used TypeScript, and so I'd love to get some feedback on this PR.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -98,7 +98,7 @@ Keep in mind that [functions are objects too](https://developer.mozilla.org/en-U
 
 ##### .numericString(value)
 
-Returns `true` for a string that represents a number, for example `'42'` and `'-8'`.
+Returns `true` for a string that represents a number. For example, `'42'` and `'-8'`.
 
 ##### .regExp(value)
 ##### .date(value)
--- a/readme.md
+++ b/readme.md
@@ -97,6 +97,9 @@ All the below methods accept a value and returns a boolean for whether the value
 Keep in mind that [functions are objects too](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions).
 
 ##### .numericString(value)
+
+Returns `true` for a string that contains a number.
+
 ##### .regExp(value)
 ##### .date(value)
 ##### .error(value)
--- a/readme.md
+++ b/readme.md
@@ -98,7 +98,7 @@ Keep in mind that [functions are objects too](https://developer.mozilla.org/en-U
 
 ##### .numericString(value)
 
-Returns `true` for a string that contains a number.
+Returns `true` for a string that represents a number (e.g. `'42'` or `'-8'`).
 
 ##### .regExp(value)
 ##### .date(value)
--- a/readme.md
+++ b/readme.md
@@ -96,6 +96,7 @@ All the below methods accept a value and returns a boolean for whether the value
 
 Keep in mind that [functions are objects too](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions).
 
+##### .numericString(value)
 ##### .regExp(value)
 ##### .date(value)
 ##### .error(value)
--- a/readme.md
+++ b/readme.md
@@ -98,7 +98,7 @@ Keep in mind that [functions are objects too](https://developer.mozilla.org/en-U
 
 ##### .numericString(value)
 
-Returns `true` for a string that represents a number (e.g. `'42'` or `'-8'`).
+Returns `true` for a string that represents a number, for example `'42'` and `'-8'`.
 
 ##### .regExp(value)
 ##### .date(value)
--- a/source/index.ts
+++ b/source/index.ts
@@ -136,8 +136,8 @@ namespace is { // tslint:disable-line:no-namespace
 	// tslint:enable:variable-name
 
 	export const nullOrUndefined = (value: any): value is null | undefined => null_(value) || undefined(value);
-	export const numericString = (value: string): boolean =>
-		!Number.isNaN(Number(value));
+	export const numericString = (value: any): boolean =>
+		string(value) && value.length > 0 && !Number.isNaN(Number(value));
 
 	export const array = Array.isArray;
 	export const buffer = isBuffer;
--- a/source/index.ts
+++ b/source/index.ts
@@ -135,10 +135,13 @@ namespace is { // tslint:disable-line:no-namespace
 	export const symbol = isOfType<Symbol>('symbol');
 	// tslint:enable:variable-name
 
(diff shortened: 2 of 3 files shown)


source/index.ts
```diff
@@ -135,10 +135,13 @@ namespace is { // tslint:disable-line:no-namespace
 	export const symbol = isOfType<Symbol>('symbol');
 	// tslint:enable:variable-name
 
+	export const nullOrUndefined = (value: any): value is null | undefined => null_(value) || undefined(value);
+	export const numericString = (value: any): value is String =>
```
gioragutt: if this is supposed to be a `numericString` how come you accept `any`? and how does it indicate that `value is String`?
First, use `string` over `String`, second - I believe this is not the correct signature for this method.
I think that it should just return bool, since you can't really create a type guard that assures you that `string` is typed as a `number`, you can just have a bool that tells you that, and you can just `Number(_)` it...
```suggestion
	export const numericString = (value: string | number): bool =>
```

the `|  number` is optional, only if you feel that's necessary...

itaisteinherz:
@gioragutt Thanks for the feedback! I've fixed the method's signature as per your suggestions.
The method currently only accepts a `string`, since accepting a `number` as well seems pointless to me.

Also, as some comments in #54 indirectly discuss the result of running `is.numericString(undefined)`, maybe we should revert the method's parameter to be `any`.

@sindresorhus What do you think?

source/index.ts
```diff
@@ -136,8 +136,8 @@ namespace is { // tslint:disable-line:no-namespace
 	// tslint:enable:variable-name
 
 	export const nullOrUndefined = (value: any): value is null | undefined => null_(value) || undefined(value);
-	export const numericString = (value: any): value is String =>
-		!Number.isNaN(Number(value)) && !nullOrUndefined(value);
+	export const numericString = (value: string): bool =>
+		!Number.isNaN(Number(value));
```
gioragutt: why did you remove the `nullOrUndefined` check?

itaisteinherz: I removed it seeing that if `value` is either `null` or `undefined`, the TS compiler would immediately return an error (since the parameter is not a `string`).
In case we revert to accepting `any`, the `nullOrUndefined` check would be required again.

gioragutt: Got it: add a `is.string` check. and then you can change `value` back to being `any`.

```ts
export const numericString = (value: any) =>
    is.string(value) && !Number.isNaN(Number(value));
```

brandon93s: I believe the intent here is to check if a variable contains any representation of a number:

```js
if (is.number(val) || is.numericString(val)) {
   console.log(`The unicorn is ${val}`)
}
```

With that I mind, I it should accept `any` and type check `string` internally.
```

File: source/tests/test.ts [hunk 3]
```diff
@@ -426,6 +426,12 @@ test('is.symbol', t => {
 	testType(t, 'symbol');
 });
 
+test('is.numericString', t => {
+	t.true(m.numericString('5'));
+	t.true(m.numericString('-3.2'));
+	t.true(m.numericString('Infinity'));
```
brandon93s: The following test case should not pass, but does:

```js
t.true(m.numericString(''));
```

```js
Number('')
// => 0
```

We need to guard against empty string.

source/index.ts
sindresorhus: Yes, it should accept `any`, like all the other checks.

source/tests/test.ts
gioragutt: Why not use `testType` here as well, instead of just checking several specific use cases?

The tests wouldn't even break here if you replaces the implementation to `return true;`.

itaisteinherz:
@brandon93s I've changed the type of `value` to `any`, fixed the issue you pointed out and added a test for it.
@gioragutt I've moved to using `testType` (I tried to use it before I even opened the PR but couldn't because there was a problem I haven't figured out until now).

gioragutt:
@itaisteinherz do you need help with working something out? We can talk over WhatsApp if you need (Israeli here as well)

readme.md
```diff
@@ -96,6 +96,7 @@ All the below methods accept a value and returns a boolean for whether the value
 
 Keep in mind that [functions are objects too](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions).
 
+##### .numericString(value)
```
sindresorhus: There should be a description here, as it's not immediately clear what a `numericString` entails.

readme.md
```diff
@@ -96,6 +96,10 @@ All the below methods accept a value and returns a boolean for whether the value
...
+
+Returns `true` for a string that contains a number.
```
sindresorhus: That's not entirely correct. It returns true for a string that **represents** a number. If it was *contains*, then this would be valid `is.numericString('foo 4 bar') //=> true`.

joker314:
This was a really good PR!

Could I just note that it's not entirely clear that the string `"NaN"` should be rejected (given that `NaN` is of type number, and `"Infinity"` and "-Infinity"` aren't. Should this be documented?

itaisteinherz:
> Could I just note that it's not entirely clear that the string `"NaN"` should be rejected (given that `NaN` is of type number, and `"Infinity"` and `"-Infinity"` aren't. Should this be documented?

I personally don't think so, since we specifically check in the source code if the given `value` is not `NaN` (see [here](https://github.com/sindresorhus/is/pull/67/files#diff-21e9ddd93162651bd36f6e5bbfca8460R140)).

joker314:
People don't read the source code of the libraries they use, normally.

Instead, they refer to documentation (that's the README.md file here). The documentation just says that it returns true for a "string representing a number".

There are different meanings of the word "number". In JavaScript, `NaN`, `10`, `Infinity` are all numbers. The function introduced in these changes considers `"Infinity"` and `"10"` numbers, but it doesn't consider `"NaN"` a number. I'm not criticising this design choice 🙂, instead, I'm asking if it's a good idea for this design choice to be noted down for developers to see.

Not putting something in the documentation just because the code is self-explanatory isn't a good idea.

```javascript
typeof NaN === 'number' // true
```

itaisteinherz:
Opinions? @sindresorhus @gioragutt @brandon93s

sindresorhus:
I agree it should be documented: https://github.com/sindresorhus/is/commit/69bbf2d4bed61c0ceea3b38f93fb01fda94f52ee

PR #68, opened 2018-10-30
Refactor `any` to `unknown` where possible

### LINKED ISSUE #62: Use Typescript 3.0 `unknown` type
With the new `unknown` type on the horizon in TS 3.0 I feel like we could refactor `any`s to use `unknown` and an explicit cast if necessary. I did not tinker with that yet, just immediately came to my mind upon reading.

https://blogs.msdn.microsoft.com/typescript/2018/07/30/announcing-typescript-3-0/

What do you think?

ltetzlaff:
Resolves #62 

It doesn't work for cases such as:

```ts
const sth(arg: unknown) => arg.prop === undefined
```

where we access members of `unknown`, keys of `WeakMap` and `WeakSet` or stuff like the methods on `Number`.

Doesn't really bring anything new to the table but maybe the `unknown` type is being worked on further or so 🤷‍♂️

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -56,10 +56,10 @@ export const enum TypeName {
 }
 
 const toString = Object.prototype.toString;
-const isOfType = <T>(type: string) => (value: any): value is T => typeof value === type;
+const isOfType = <T>(type: string) => (value: unknown): value is T => typeof value === type;
 const isBuffer = (input: any): input is Buffer => !is.nullOrUndefined(input) && !is.nullOrUndefined(input.constructor) && is.function_(input.constructor.isBuffer) && input.constructor.isBuffer(input);
 
-const getObjectType = (value: any): TypeName | null => {
+const getObjectType = (value: unknown): TypeName | null => {
 	const objectName = toString.call(value).slice(8, -1) as string;
 
 	if (objectName) {
@@ -69,9 +69,9 @@ const getObjectType = (value: any): TypeName | null => {
 	return null;
 };
 
-const isObjectOfType = <T>(type: TypeName) => (value: any): value is T => getObjectType(value) === type;
+const isObjectOfType = <T>(type: TypeName) => (value: unknown): value is T => getObjectType(value) === type;
 
-function is(value: any): TypeName { // tslint:disable-line:only-arrow-functions
+function is(value: unknown): TypeName { // tslint:disable-line:only-arrow-functions
 	switch (value) {
 		case null:
 			return TypeName.null;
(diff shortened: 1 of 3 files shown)


ltetzlaff:
> Wouldn't it better to still use unknown and just use as to define the exact type where we need to access the members?

True, albeit not being necessarily beautiful it gives us cleaner interfaces that are all typed `unknown` now.

source/index.ts
```diff
@@ -57,7 +57,7 @@ export const enum TypeName {
 
 const toString = Object.prototype.toString;
 const isOfType = <T>(type: string) => (value: unknown): value is T => typeof value === type;
-const isBuffer = (input: any): input is Buffer => !is.nullOrUndefined(input) && !is.nullOrUndefined(input.constructor) && is.function_(input.constructor.isBuffer) && input.constructor.isBuffer(input);
+const isBuffer = (input: unknown): input is Buffer => !is.nullOrUndefined(input) && !is.nullOrUndefined((input as Buffer).constructor) && is.function_((input as Buffer).constructor.isBuffer) && (input as Buffer).constructor.isBuffer(input);
```
sindresorhus: Why do we need the `as Buffer` when we already have the `input is Buffer` type guard?

source/index.ts
```diff
@@ -258,11 +258,11 @@ namespace is { // tslint:disable-line:no-namespace
...
 	];
 
-	export const domElement = (value: any): value is DomElement => object(value) as unknown && value.nodeType === NODE_TYPE_ELEMENT && string(value.nodeName) &&
-		!plainObject(value) && DOM_PROPERTIES_TO_CHECK.every(property => property in value);
+	export const domElement = (value: unknown): value is DomElement => object(value) && (value as DomElement).nodeType === NODE_TYPE_ELEMENT && string((value as DomElement).nodeName) &&
+		!plainObject(value) && DOM_PROPERTIES_TO_CHECK.every(property => property in (value as DomElement));
 
-	export const observable = (value: any) => Boolean(value && value[symbolObservable] && value === value[symbolObservable]());
-	export const nodeStream = (value: any): value is NodeStream => !nullOrUndefined(value) && isObject(value) as unknown && function_(value.pipe) && !observable(value);
+	export const observable = (value: unknown) => Boolean(value && (value as any)[symbolObservable] && value === (value as any)[symbolObservable]());
```
sindresorhus: Could we use `Object` instead of `any` here?

source/index.ts
ltetzlaff: Well the typeguard `is Buffer` is more like a return type, so the code placed *after* the invocation `.isBuffer(input)` will know if `input` is a `Buffer` or not but within the body this is just being evaluted. Also we're using custom typeguards thus sort of overriding the default Typescript behavior (helping by explicitly stating what it can't find out on its own 🙏)

sindresorhus: Good point 👍

source/index.ts
ltetzlaff: We are lacking the index signature there on `Object`, `{}`, `object` etc.

I also can't wrap my head around what type to cast it to, attempting to type it out with a `symbolObservable` as index signature doesn't work, because types of index signatures have to be `string | number` thus it gives me:

```
[ts] Type 'symbol' cannot be used as an index type.
```

That's why I settled for any 🤷‍♂️👍

sindresorhus: Ah I see. Alright `any` is fine then.

sindresorhus:
Thank you! :)

PR #69, opened 2018-11-02
Minor documentational correction (return -> returns)

joker314:
When the subject is singular, regular verbs tend to have an additional 's'. Hence, 'return' is modified to 'returns' to implement this rule. With it, the text may flow more easily to the reader


sindresorhus:
Thanks for fixing my bad grammar :)

PR #74, opened 2018-12-12
Improve the `is.observable` check

### LINKED ISSUE #72: Importing this module mutates global Symbol.observable
This module imports `symbol-observable`, which patches global `Symbol.observable`.
I just spent half an hour tracking down a bug which was caused by this - rxjs uses and checks for the presence of either `Symbol.observable` (if available) or of `"@@observable"`. If RxJS was imported first and some Observable created, it would receive the `"@@observable"` shim. If then this module is imported, it mutates `Symbol.observable`. If then some dependency imports a different version of RxJS, it receives the `Symbol.observable`, and Rx `from()` is unable to recognize the Observable as an Observable.

sindresorhus:
Fixes #72

I ended up going with the solution in https://github.com/sindresorhus/is/issues/72#issuecomment-444739348, as the solution in https://github.com/sindresorhus/is/issues/72#issuecomment-444738749 did not work correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,9 +43,6 @@
 		"verify",
 		"compare"
 	],
-	"dependencies": {
-		"symbol-observable": "^1.2.0"
-	},
 	"devDependencies": {
 		"@sindresorhus/tsconfig": "^0.1.0",
 		"@types/jsdom": "^11.12.0",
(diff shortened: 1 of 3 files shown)


sindresorhus:
@felixfbecker Can you confirm this fixes your problem?

Issue #70, opened 2018-11-02
Fix `.even()`/`.odd()` checks


sindresorhus:
I think those method names should be more explicit. Currently, they check if a number is an integer and whether it's even or odd. The problem is that a float can also be even or odd.

We should either:

1. Rename them to `.evenInteger()` / `.oddInteger()`.
2. Remove the integer check and rename them to `.evenNumber()` / `.oddNumber()`.

I'm leaning towards the former as I don't really see the use-case for floats for this, but I'm open to feedback.

brandon93s:
> The problem is that a float can also be even or odd.

Are you referring to `2.00` or `2.02`? I would expect `2.00` to return `true` and `2.02` to return `false`. This is the current behavior since `Number.isInteger(2.00)` is `true` for zero decimal values.

sindresorhus:
@brandon93s The problem is that no one on the internet can agree on the definition of an odd float. Many articles say that an even number is an integer that can be divided by 2 and results in no remainders, and that an odd number is something that is *not* an even number. But by that definition, `2.02` is a valid odd number.

It would just be good to make the methods explicit as there is no universal definition of what to expect when you supply a float to even()/odd() checks.

And yes, I agree `2.00` is an integer in JS terms.

sindresorhus:
So I think we should do this:

> Rename them to .evenInteger() / .oddInteger().

PR #77, opened 2019-01-12
#70: renamed odd to oddInteger and even to evenInteger for clarity

scottenriquez:


sindresorhus:
Would you be able to fix the merge conflict?

PR #78, opened 2019-01-26
Stop using TypeScript `namespace`

sindresorhus:
~~This enables to use `is.null`, `is.class` and `is.function` in TypeScript without the underscore suffix. 🎉~~ Nope...

---

> TypeScript 3.1 brings the ability to define properties on function declarations and const-declared functons, simply by assigning to properties on these functions in the same scope. This allows us to write canonical JavaScript code without resorting to namespace hacks.
> \- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#properties-declarations-on-functions

---

// @ltetzlaff @SamVerschueren

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -58,7 +58,7 @@ export const enum TypeName {
 
 const toString = Object.prototype.toString;
 const isOfType = <T>(type: string) => (value: unknown): value is T => typeof value === type;
-const isBuffer = (input: unknown): input is Buffer => !is.nullOrUndefined(input) && !is.nullOrUndefined((input as Buffer).constructor) && is.function((input as Buffer).constructor.isBuffer) && (input as Buffer).constructor.isBuffer(input);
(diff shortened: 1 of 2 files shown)


gioragutt:
How does the `.d.ts` file look like after this?
When thinking about migrating `is` to `ts`, this was basically my biggest question.
I wanted the code to look like what you've done in the PR, but couldn't tell how would the typings for the `is` object would look like.

sindresorhus:
Ugh, turns out TypeScript still creates a `namespace` for the generated TypeScript definition and there it uses `function` without the underscore too, which is reserved in TypeScript, which means TypeScript generates an invalid TypeScript definition…

I tried the changes out in [`got`](https://github.com/sindresorhus/got) and got the error:

```
source/utils/is-form-data.ts(3,70): error TS2339: Property 'function' does not exist on type 'typeof is'.
```

ltetzlaff:
Oof, that's nasty 🙈

sindresorhus:
Here's the diff of new the new TS definition file: https://gist.github.com/sindresorhus/5db5b16c30e8539af3f7d0983b6886ed/revisions#diff-6412df35290ec1e8106ec1768717873f

sindresorhus:
And here I actually started liking TypeScript...

sindresorhus:
I'm still gonna merge this as I don't really like `namespace` and it's nice to stay as close to vanilla JS as possible.

PR #80, opened 2019-02-01
Improve the TypeScript types

sindresorhus:
Mostly more use of type guards.

Initially triggered by https://github.com/sindresorhus/got/pull/712#discussion_r252549601, but I found a lot more when fixing that.

---

// @ltetzlaff @fnky

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -8,7 +8,9 @@
 const URLGlobal = typeof URL === 'undefined' ? require('url').URL : URL;
 
 type TypedArray = Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array | Int32Array | Uint32Array | Float32Array | Float64Array;
-type Primitive = null | undefined | string | number | boolean | Symbol;
+
+// TODO: This should be able to be `not object` when the `not` operator is out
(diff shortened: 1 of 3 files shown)


source/index.ts
```diff
@@ -296,30 +303,34 @@ is.nodeStream = (value: unknown): value is NodeStream => !is.nullOrUndefined(val
...
 is.even = isAbsoluteMod2(0);
 is.odd = isAbsoluteMod2(1);
 
-const isWhiteSpaceString = (value: unknown) => is.string(value) && /\S/.test(value) === false;
+is.emptyArray = (value: unknown): value is never[] => is.array(value) && value.length === 0;
+is.nonEmptyArray = (value: unknown): value is unknown[] => is.array(value) && value.length > 0;
 
-is.emptyArray = (value: unknown) => is.array(value) && value.length === 0;
-is.nonEmptyArray = (value: unknown) => is.array(value) && value.length > 0;
+is.emptyString = (value: unknown): value is string => is.string(value) && value.length === 0;
```
ltetzlaff: we can even go for

```ts
value is ""
```

here :+1:

sindresorhus: That is very clever! Didn't realize that would work.

PR #86, opened 2019-04-19
Test coverage for `is(value)`

### LINKED ISSUE #7: is(value) is not slightly tested
There are no tests that `is(value)` works correctly.

This makes it hard to make changes for it, f.e replace all the primitive if checks with
```typescript
if (is.primitive(value)) {
   return typeof value;
}
```

I had an idea regarding the `testType` method.
Currently, it accepts an optional argument that is an array (excludes).

I played with the method a bit, and it looks something like this:
```typescript
const testTypeName = (t, fixture, typename) => {
	t.is(m(fixture), typename, `Type name of ${util.inspect(fixture)} is not ${typename}`);
};

// This ensure a certain method matches only the types
// it's supposed to and none of the other methods' types
const testType = (t, type, {exclude, typename} = {}) => {
	for (const [key, value] of types) {
		if (exclude && exclude.indexOf(key) !== -1) {
			continue;
		}

		const assert = key === type ? t.true.bind(t) : t.false.bind(t);
		const is = m[type];
		const fixtures = Array.isArray(value) ? value : [value];

		for (const fixture of fixtures) {
			assert(is(fixture), `Value ${util.inspect(fixture)} is not ${type}`);

			if (typename) {
				testTypeName(t, fixture, typename);
			}
		}
	}
};
```

basically the third parameter becomes an object with which we can pass optional parameters.

tests with excludes now look like:

```typescript
test('is.undefined', t => {
	testType(t, 'undefined', {
		exclude: ['nullOrUndefined'],
		typename: 'undefined'
	});
});

test('is.null', t => {
	testType(t, 'null', {
		exclude: ['nullOrUndefined'],
		typename: 'null'
	});
});
```

See that I can easily pass data(`typename`) for checking `is(null) === 'null')`, and just not pass it if I don't want to check it.

This also allows the tests to be a little more verbose (you explicitly say that you **exclude** those types from the test).

MrHen:
Fixes #7.

I chose a slightly different path than the one suggested in the issue. There were only two minor annoyances that you might want cleaned up / tweaked:

* For some reason, `TypeName.GeneratorFunction` has a value of `'GeneratorFunction'` but the value returned from `is` is `'Generator'`. If the actual string of `'GeneratorFunction'` is not being used, could we change it to `'Generator'`?
* The various HTML elements all have different types that are not included in `TypeName`. They were also mixed into a single test type of `domElement` which meant that the `typename` helper would not work properly. I don't think anything needs to be changed here; just calling it out.

### DIFF
--- a/test/test.ts
+++ b/test/test.ts
@@ -801,12 +801,19 @@ test('is.domElement', t => {
 	testType(t, 'domElement');
 	t.false(is.domElement({nodeType: 1, nodeName: 'div'}));
 
-	t.is(is(createDomElement('div')), 'HTMLDivElement');
-	t.is(is(createDomElement('input')), 'HTMLInputElement');
-	t.is(is(createDomElement('span')), 'HTMLSpanElement');
-	t.is(is(createDomElement('img')), 'HTMLImageElement');
-	t.is(is(createDomElement('canvas')), 'HTMLCanvasElement');
-	t.is(is(createDomElement('script')), 'HTMLScriptElement');
+	const htmlTagNameToTypeName = {
+		div: 'HTMLDivElement',
+		input: 'HTMLInputElement',
+		span: 'HTMLSpanElement',
+		img: 'HTMLImageElement',
+		canvas: 'HTMLCanvasElement',
+		script: 'HTMLScriptElement'
+	};
+
+	Object.entries(htmlTagNameToTypeName).forEach(([tagName, typeName]) => {
+		const domElement = createDomElement(tagName);
+		t.is(is(domElement), typeName);
+	});
 });
 
 test('is.observable', t => {
--- a/test/test.ts
+++ b/test/test.ts
@@ -808,10 +808,10 @@ test('is.domElement', t => {
 		script: 'HTMLScriptElement'
 	};
 
-	Object.entries(htmlTagNameToTypeName).forEach(([tagName, typeName]) => {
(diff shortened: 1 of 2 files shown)


test/test.ts
```diff
@@ -753,6 +800,13 @@ test('is.inRange', t => {
 test('is.domElement', t => {
 	testType(t, 'domElement');
 	t.false(is.domElement({nodeType: 1, nodeName: 'div'}));
+
+	t.is(is(createDomElement('div')), 'HTMLDivElement');
```
gioragutt: I'd refactor this to something like:
```typescript
const htmlTagNameToTypeName = {
    span: 'HTMLSpanElement',
    ...
}

Object.entries(htmlTagNameToTypeName).forEach(([tagName, typeName]: string[]) => {
    const domElement = createDomElement(tagName);
    t.is(is(domElement), typeName);
})
```

test/test.ts
```diff
@@ -392,21 +441,19 @@ const testType = (t: ExecutionContext, type: string, exclude?: string[]) => {
 			continue;
 		}
 
-		const assert = key === type ? t.true.bind(t) : t.false.bind(t);
+		const isTypeUnderTest = key === type;
```
gioragutt: good variable name, +1

MrHen:
Turns out `GeneratorFunction` _does_ need to be `'GeneratorFunction'` so I just added `Generator` as its own entry in the enum.

sindresorhus:
🙌

PR #87, opened 2019-05-02
Add support for BigInt

### LINKED ISSUE #39: Support BigInt detection
http://2ality.com/2017/03/es-integer.html
https://groups.google.com/forum/#!msg/v8-dev/x571Gr0khNo/y8Jk0_vSBAAJ

Also `BigInt64Array` and `BigUint64Array`

Blocked by https://github.com/Microsoft/TypeScript/issues/15096

vladfrangu:
Fixes #39 

- [x] Implements support for the BigInt, BigInt64Array and BigUint64Array

- [x] All tests pass

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
 language: node_js
 node_js:
   - '10'
-  - '8'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
 language: node_js
 node_js:
   - '10'
+  - '8'
--- a/test/test.ts
+++ b/test/test.ts
@@ -2,6 +2,8 @@ import fs from 'fs';
 import net from 'net';
 import Stream from 'stream';
 import util from 'util';
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const URLGlobal = typeof URL === 'undefined' ? require('url').URL : URL;
 import tempy from 'tempy';
 import test, {ExecutionContext} from 'ava';
 import {JSDOM} from 'jsdom';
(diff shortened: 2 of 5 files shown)


sindresorhus:
Why are you removing Node.js 8 support? That seems unrelated.

vladfrangu:
I thought nodejs 8 didn't have bigint support. Let me check real quick...

vladfrangu:
image

nodejs 8 does NOT have BigInt support. (quite a shame too)

sindresorhus:
We're not actually *using* BigInt, we're just detecting it, so we can still preserve support for Node.js 8. You can just comment out the tests for now.

tsconfig.json
```diff
@@ -2,10 +2,10 @@
 	"extends": "@sindresorhus/tsconfig",
 	"compilerOptions": {
 		"outDir": "dist",
-		"target": "es2017",
+		"target": "esnext",
```
sindresorhus: We don't have to change `target`. TS doesn't polyfill BigInt anyway.

tsconfig.json
```diff
@@ -2,10 +2,10 @@
...
 		"lib": [
 			"es2017",
```
sindresorhus: This can be removed now as `esnext` includes all previous versions.

vladfrangu:
Yep, after reading your review, I realized you only did typeof / toString calls and checks, which would mean nodejs 8 support would still be present! I've done your requested changes now, as well as commented out all tests that only work on nodejs >=10. Hope this helps!

test/test.ts
```diff
@@ -2,7 +2,6 @@ import fs from 'fs';
 import net from 'net';
 import Stream from 'stream';
 import util from 'util';
-import {URL} from 'url';
```
sindresorhus: This too

.travis.yml
```diff
@@ -1,4 +1,3 @@
 language: node_js
 node_js:
   - '10'
-  - '8'
```
sindresorhus: And this

vladfrangu:
@sindresorhus should be done now! If there's any other changes you'd like me to do, please let me know 👍

sindresorhus:
Nice work 👌

PR #89, opened 2019-05-16
Drop non-working `default` export for TypeScript users

### LINKED ISSUE #88: Runtime errors if compiled with `"esModuleInterop": false`
Thanks for the lovely library!  With version `0.17.0`, we noticed that our tests suddenly weren't passing:
https://github.com/googleapis/nodejs-pubsub/pull/591

Using TypeScript, everything compiles fine - then we'd get this at runtime:

```
const iss = is_1.default.string('👋');
                         ^

TypeError: Cannot read property 'string' of undefined
    at Object.<anonymous> (/Users/beckwith/Code/isbroke/build/src/index.js:4:26)
    at Module._compile (internal/modules/cjs/loader.js:816:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:827:10)
    at Module.load (internal/modules/cjs/loader.js:685:32)
    at Function.Module._load (internal/modules/cjs/loader.js:620:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:877:12)
    at internal/main/run_main_module.js:21:11
```

I have a minimal repo here:
https://github.com/JustinBeckwith/is-broke

Just clone, `npm install`, and `npm start` to see the error.  One thing to note - we do not have `esModuleInterop` enabled in our tsconfig.  If I set that to `true` - everything does seem to work.  

We found that setting `esModuleInterop` in our modules could have downstream effects for other users (like this case), so we keep it off. I noticed that your base level `tsconfig` enables both synthetic default imports and es module interop:
https://github.com/sindresorhus/tsconfig/blob/master/tsconfig.json#L9

TL;DR: I *think* the answer here is to disable `esModuleInterop` as a default configuration for your TypeScript modules. Could be wrong 🤷‍♂

bfred-it:
Fixes #88 

I kept the TODO like in https://github.com/sindresorhus/mem/blob/10f13c0904e497a35bb4cba907490aad1ba6d879/index.d.ts#L93

### DIFF
--- a/test/test.ts
+++ b/test/test.ts
@@ -7,10 +7,11 @@ import test, {ExecutionContext} from 'ava';
 import {JSDOM} from 'jsdom';
 import {Subject, Observable} from 'rxjs';
 import ZenObservable from 'zen-observable';
-import is, {TypeName} = require('../source');
+import is = require('../source');
 
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 const URLGlobal = typeof URL === 'undefined' ? require('url').URL : URL;
+const {TypeName} = is;
 
 const isNode10orHigher = Number(process.versions.node.split('.')[0]) >= 10;
 
--- a/test/test.ts
+++ b/test/test.ts
@@ -7,7 +7,7 @@ import test, {ExecutionContext} from 'ava';
 import {JSDOM} from 'jsdom';
 import {Subject, Observable} from 'rxjs';
 import ZenObservable from 'zen-observable';
-import is, {TypeName} from '../source';
+import is, {TypeName} = require('../source');
 
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 const URLGlobal = typeof URL === 'undefined' ? require('url').URL : URL;
--- a/test/test.ts
+++ b/test/test.ts
@@ -7,10 +7,13 @@ import test, {ExecutionContext} from 'ava';
 import {JSDOM} from 'jsdom';
 import {Subject, Observable} from 'rxjs';
 import ZenObservable from 'zen-observable';
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
 import is = require('../source');
 
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 const URLGlobal = typeof URL === 'undefined' ? require('url').URL : URL;
+
 const {TypeName} = is;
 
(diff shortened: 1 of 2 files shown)


bfred-it:
I'm not entirely sure of how to fool [import/default](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/default.js) without having an actual `export default`

See eslint's log at https://travis-ci.org/sindresorhus/is/jobs/533220117

sindresorhus:
I would prefer just reverting back to:

```ts
export default is;

// For CommonJS default export support
module.exports = is;
module.exports.default = is;
```

Which we know are working.

bfred-it:
`export default is` isn't right? https://github.com/sindresorhus/mem/issues/31

sindresorhus:
Not exactly, since the source here is TypeScript, while https://github.com/sindresorhus/mem/issues/31 is about CommonJS.

bfred-it:
The previous change broke compatibility (we’re still v0.x) but given the TODO comment in that PR I think your modules aren’t meant to export `default` in the first place. `is` is still a CJS module and thus it should not be imported via `import is from x` in TypeScript, especially without esInterop enabled 🤷‍♂️

sindresorhus:
You have to update the test to use `import is = require('.')` syntax.

test/test.ts
```diff
@@ -7,11 +7,15 @@ import test, {ExecutionContext} from 'ava';
 import {JSDOM} from 'jsdom';
 import {Subject, Observable} from 'rxjs';
 import ZenObservable from 'zen-observable';
-import is, {TypeName} from '../source';
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
```
bfred-it: Either `import/default` follows `esModuleInterop` and therefore allows `import X from Y` on CJS modules, or `@typescript-eslint/no-require-imports` should be disabled.

Since I have to disable a rule, I might as well leave the old ESM import and disable `import/default` on that line

bfred-it:
The last two commits are valid but there's an unrelated TS issue:

```
  Uncaught exception in test/test.ts
  test/test.ts:30:13 - error TS2749: 'TypeName' refers to a value, but is being used as a type here.
  
  30  typename?: TypeName;
                 ~~~~~~~~
```

sindresorhus:
Fixing this properly requires too many breaking changes. We'll also have to put every type export behind a namespace. I'm just gonna revert the original commit and keep it like this until we can move to ESM later this year.

PR #90, opened 2019-05-31
Make `is.number(NaN)` return false

### LINKED ISSUE #63: Any value in `is.numericValue` or similar?
At the moment, `is.number` returns `true` if the passed value is `NaN`. I'd quite like a function that returned `true` iff the passed value was a number with a real numeric value (`0`, `-0`, `+0`, etc. still count, `Infinity` I'd take either way (arguably `true` as Infinity is a real numeric value, just not actually countable)).

Not having yet thought about any edge cases, I imagine this would be a simple function:

```js
export const numericValue = (value) => number(value) && ! nan(value)
```

kainiedziela:
Fixes #63 

> I'm been thinking about this a lot, and I'm now leaning towards just changing `is.number()` to not accept `NaN`. It can work if we clearly document that it does not accept `NaN`. Just because `typeof` is broken, doesn't mean that we can't do better. This package is all about creating the most user-friendly type checks.

And so, `is.number(NaN)` returns `false`. No further changes, so `-Infinite` and `Infinite` and `-0` still return `true`.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -86,7 +86,7 @@ Keep in mind that [functions are objects too](https://developer.mozilla.org/en-U
 
 ##### .numericString(value)
 
-Returns `true` for a string that represents a number satisfying `is.number`.
+Returns `true` for a string that represents a number satisfying `is.number`, for example, `'42'` and `'-8.3'`.
 
 Note: `'NaN'` returns `false`, but `'Infinity'` and `'-Infinity'` return `true`.
 
--- a/readme.md
+++ b/readme.md
@@ -69,7 +69,7 @@ All the below methods accept a value and returns a boolean for whether the value
 ##### .string(value)
 ##### .number(value)
 
-Note: `is.number(NaN)` returns `false`.
+Note: `is.number(NaN)` returns `false`. This intentionally deviates from `typeof` behavior to increase user-friendliness of `is` type checks.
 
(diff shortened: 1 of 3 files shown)


source/index.ts
```diff
@@ -118,7 +118,7 @@ const isObject = (value: unknown): value is object => typeof value === 'object';
 
 is.undefined = isOfType<undefined>('undefined');
 is.string = isOfType<string>('string');
-is.number = isOfType<number>('number');
+is.number = (value: unknown): value is number => isOfType<number>('number')(value) && !is.nan(value);
```
gioragutt: Extract isTypeOf(‘number’) to a constant before ‘is.number’.  No need to create it over and over

readme.md
```diff
@@ -68,6 +68,9 @@ All the below methods accept a value and returns a boolean for whether the value
 ##### .null(value)
 ##### .string(value)
 ##### .number(value)
+
+Note: `is.number(NaN)` returns `false`.
```
sindresorhus: This should also describe the reasoning for why it return false.

sindresorhus:
> do not accept NaN in is.number()

The PR title is incorrect. It does accept NaN, it just doesn't return true.

sindresorhus:
Thanks :)

Issue #51, opened 2018-05-18
Encountered compilation errors related to SharedArrayBuffer


DominicKramer:
When using `@sindresorhus/is@^0.9.0`, I am encountering the following error message when compiling TypeScript code that imports the module:
```
[...]$ tsc
node_modules/@sindresorhus/is/dist/index.d.ts:81:55 - error TS2304: Cannot find name 'SharedArrayBuffer'.

81     const sharedArrayBuffer: (value: any) => value is SharedArrayBuffer;
```
The following are all of the relevant files in the workspace.

**package.json**
```json
{
  "name": "temp",
  "version": "1.0.0",
  "license": "Apache-2.0",
  "devDependencies": {
    "@types/node": "^10.1.2",
    "typescript": "^2.8.3"
  },
  "dependencies": {
    "@sindresorhus/is": "^0.9.0"
  }
}
```

**tsconfig.json**
```json
{
  "compilerOptions": {
    "allowSyntheticDefaultImports": true,
    "allowUnreachableCode": false,
    "allowUnusedLabels": false,
    "declaration": true,
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "lib": ["es2016"],
    "module": "commonjs",
    "noEmitOnError": true,
    "noFallthroughCasesInSwitch": true,
    "noImplicitReturns": true,
    "pretty": true,
    "sourceMap": true,
    "strict": true,
    "target": "es2016",
    "rootDir": ".",
    "outDir": "build"
  },
  "include": [
    "src/*.ts",
    "src/**/*.ts",
    "test/*.ts",
    "test/**/*.ts"
  ],
  "exclude": [
    "node_modules"
  ]
}
```
**src/index.ts**
```ts
import is from '@sindresorhus/is';
```

sindresorhus:
We have `lib.es2017.sharedmemory` in [our config](https://github.com/sindresorhus/is/blob/37f6cc3fe3df89e3b77aa9bb2cb87957fa9759ce/tsconfig.json#L14), but I didn't think that would be required for consumers too. It's weird that TS handles it like this, unless I'm missing something.

@SamVerschueren Any ideas?

SamVerschueren:
You can set `skipLibCheck` to `false`. See https://www.typescriptlang.org/docs/handbook/compiler-options.html. By default, TS checks all shipped type definitions in node_modules as well. But I agree that it's weird for consumers to care about this. Will dive deeper into it when I have some more time.

[...]

PR #92, opened 2019-08-25
Fix unnecessary global inclusion of dom lib

AaronFriel:
TypeScript packages depending on this and importing `@sindresorhus/is`
end up with the lib.dom.d.ts in global scope thanks to the deleted
lines.

That's almost never what any user of the package would want, and it has
the side effect of putting a slew of common variable names (including
the variable names "event" and "name" and "parent") into a global scope.

It's particularly sinister because on many of these, TypeScript won't
complain, for example "name" is of type never.

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -1,6 +1,3 @@
-/// <reference lib="esnext"/>
-/// <reference lib="dom"/>
-
 // TODO: Use the `URL` global when targeting Node.js 10
 const URLGlobal = typeof URL === 'undefined' ? require('url').URL : URL;
 


sindresorhus:
You have not explained why you think those lines are not needed, because they were added for a good reason.

AaronFriel:
This might be worthy of a TypeScript bug report, but the problem is that this library poisons the global scope with everything from `lib.dom.d.ts`, even in packages that don't run in a browser and thus _do not_ have those names in scope.

So I think the right way to address #51 is that downstream users should add to their `tsconfig`:

```json
{
  "compilerOptions": {
    "skipLibCheck": true
  }
}
```

That may be a controversial opinion, but it doesn't affect the usage of complex types in libraries, and it makes API breakage a "pay for what you use" model.

Please check out this repository for examples: https://github.com/AaronFriel/is-test

On the HEAD commit, `8a7c77c (HEAD -> master) Demonstrate original is library global name poisoning`, `npm run build` succeeds but `npm run start` fails at run time:

```shell
> npm run start

> is-test@1.0.0 start C:\c\af\is-test
> node ./build/src/index.js

C:\c\af\is-test\build\src\index.js:13
                ^

ReferenceError: name is not defined
    at main (C:\c\af\is-test\build\src\index.js:13:17)
    at Object.<anonymous> (C:\c\af\is-test\build\src\index.js:15:1)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:743:3)
```

On the initial commit, `ebda52c (origin/master) Demonstrate pay-for-what-you-use skipLibCheck option and correct errors`, `npm run build` fails, because the name is rightly out of scope.

Maybe this is why it should be a TypeScript bug report, because there probably ought to be a way to refer to the _types_ of `lib.dom.d.ts`, without poisoning the global scope.

AaronFriel:
Although maybe this is the best argument? 😄 

https://github.com/sindresorhus/tsconfig/blob/f329c2032b9dd8d69fdb157215db275f242a64d3/tsconfig.json#L23

sindresorhus:
> Maybe this is why it should be a TypeScript bug report, because there probably ought to be a way to refer to the types of lib.dom.d.ts, without poisoning the global scope.

Yes, please open a TypeScript issue about this and comment the link here.

If you open that TS issue, I'm ok with copy-pasting the DOM types we need here so we don't have to depend on `lib.dom.d.ts`.

> So I think the right way to address #51 is that downstream users should add to their tsconfig:

I cannot assume what people have in their tsconfig. That would make the situation much worse than now.

> Although maybe this is the best argument? 😄

No. That was done for performance reasons.

PR #97, opened 2019-11-08
Add assertion type guards

### LINKED ISSUE #91: Add "asserts" variants of type guards
A new typescript language feature is slated for the [v3.7.0 milestone](https://github.com/microsoft/TypeScript/milestone/98): `asserts`. I think adding new [`asserts value is T`](https://github.com/microsoft/TypeScript/pull/32695) variants of the type guard-enabled functions in `is` would be very useful. It would reduce the often repetitive "if-not-typeof-throw" pattern to a single line, while narrowing types.

Example implementation:

```typescript
is.assertedString = (value: unknown): asserts value is string {
  if (!is.string(value)) {
    throw new TypeError("Not a string.");
  }
}
```

<details>Without asserts:

```typescript
function f(input: unknown) {
  if (!is.string(input)) {
    throw new TypeError("Not a string.");
  }
  // input has type string here.
}
```
</details>

With asserts:

```typescript
function f(input: unknown) {
  is.assertedString(input);
  // input has type string here.
}
```

Function naming to be discussed.

joelpurra:
Uses the typescript v3.7 feature `asserts value is T` to create `assert` variants of the `is` type guards. The assertions are used to narrow types at compile time, and to throw `TypeError` at runtime for values which are not of the correct type.

```typescript
import {assert} from '@sindresorhus/is';

assert.string(foo);
```

- Each method in `is` is wrapped and mirrored in `assert`.
- Tests for `is` are duplicated for `assert`.

Notes

- The explicit typing in `interface Assert` is required for typescript to acknowledge the assertions.
- Due to the assertions requiring explicit typing, using ` property for is.assert.string()` (and so on) would require using `namespace is`, which was removed in #78. This also means that `assert` needs to be exported separately.
- Custom descriptions are used to enhance some assertion error messages. The value is not included in the error message.
- Could perhaps use Node.js' fitting `AssertionError` on the server-side, but it would require an `import`.

Fixes #91.

See

- https://github.com/sindresorhus/is/issues/91
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions
- https://github.com/microsoft/TypeScript/pull/32695


aleclarson:
Any reason why type assertions shouldn't have their own package?
```ts
import assert from '@sindresorhus/assert'
```
I think most TypeScript programs don't need them.

joelpurra:
Due to typescript's syntax changes/additions in v3.7 (such as `asserts`) this pull request originally failed `xo` linting. Upgraded typescript/linting dependencies (and rebased the `asserts` on top) to fix that; it's now confirmed to work. Will submit a separate pull request to keep the changes apart.

joelpurra:
Awaiting #101; should be rebased on top of `master` after merge.

sindresorhus:
> The value is not included in the error message.

I think it would be useful to include the value.

Maybe something like:

```
Expected value to be `number`, got `foo` (string)
```

?

readme.md
```diff
@@ -37,6 +38,15 @@ is.number(6);
 //=> true
 ```
 
+Assertions perform the same type checks, but throw errors if the type does not match.
```
sindresorhus:
```suggestion
[Assertions](#type-assertions) perform the same type checks, but throw errors if the type does not match.
```

readme.md
```diff
@@ -37,6 +38,15 @@ is.number(6);
...
+
+```js
+const {assert} = require('@sindresorhus/is');
+
+assert.string(foo);
+//=> foo is known to be a string
```
sindresorhus: `//=>` is usually used to show output, so I think the text should just be a normal comment above.

source/index.ts
```diff
@@ -385,6 +385,231 @@ const predicateOnArray = (method: ArrayMethod, predicate: Predicate, values: unk
...
+	promise: <T = unknown>(value: unknown) => asserts value is Promise<T>;
+	generatorFunction: (value: unknown) => asserts value is GeneratorFunction;
+	// eslint-disable-next-line @typescript-eslint/ban-types
+	asyncFunction: (value: unknown) => asserts value is Function;
+	// eslint-disable-next-line @typescript-eslint/ban-types
+	boundFunction: (value: unknown) => asserts value is Function;
+	regExp: (value: unknown) => asserts value is RegExp;
+	date: (value: unknown) => asserts value is Date;
+	error: (value: unknown) => asserts value is Error;
+	map: <TKey = unknown, TValue = unknown>(value: unknown) => asserts value is Map<TKey, TValue>;
```
sindresorhus:
```suggestion
	map: <Key = unknown, Value = unknown>(value: unknown) => asserts value is Map<Key, Value>;
```

sindresorhus: What do you think about adding generic type parameters to some of the `is` methods too?

source/index.ts
```diff
@@ -385,6 +385,231 @@ const predicateOnArray = (method: ArrayMethod, predicate: Predicate, values: unk
...
+	bigUint64Array: (value: unknown) => asserts value is BigUint64Array;
+	arrayBuffer: (value: unknown) => asserts value is ArrayBuffer;
+	sharedArrayBuffer: (value: unknown) => asserts value is SharedArrayBuffer;
+	dataView: (value: unknown) => asserts value is DataView;
+	urlInstance: (value: unknown) => asserts value is URL;
+	urlString: (value: unknown) => asserts value is string;
+	truthy: (value: unknown) => asserts value is unknown;
+	falsy: (value: unknown) => asserts value is unknown;
+	nan: (value: unknown) => asserts value is unknown;
+	primitive: (value: unknown) => asserts value is unknown;
```
sindresorhus:
```suggestion
	primitive: (value: unknown) => asserts value is Primitive;
```

?

joelpurra:
> > The value is not included in the error message.
> 
> I think it would be useful to include the value.
> 
> Maybe something like:
> 
> ```
> Expected value to be `number`, got `foo` (string)
> ```
> 
> ?

Useful, yes. Mostly avoided because outputting arbitrary (and in this case _provably unexpected_) input is scary. Can be seen as a security and information leakage issue. Could add something like `received value of type ${is(value)}` to add a hint though. Would that work?

source/index.ts
joelpurra: Would be nice to propagate some more typings for the cases when typescript already knows about them. Separate pull request, right?

joelpurra:
Force-pushed to remove code from #101 already in `master`, and WIP review fixes to be squashed upon approval.

sindresorhus:
> Could add something like received value of type ${is(value)} to add a hint though. Would that work?

👍

sindresorhus:
No need to force push. It makes it easier to review if you add new commits. I will squash on merge anyway.

source/index.ts
sindresorhus: > Separate pull request, right?

Yes, would be great :)

sindresorhus:
This looks good to me. Thank you  🙌

PR #100, opened 2019-12-27
Add `.asyncGenerator` and `.asyncGeneratorFunction` detection

### LINKED ISSUE #99: Add `.asyncGenerator` detection
https://jakearchibald.com/2017/async-iterators-and-generators/

forresst:
Fixes #99

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -460,9 +460,11 @@ interface Assert {
 	iterable: <T = unknown>(value: unknown) => asserts value is Iterable<T>;
 	asyncIterable: <T = unknown>(value: unknown) => asserts value is AsyncIterable<T>;
 	generator: (value: unknown) => asserts value is Generator;
+	asyncGenerator: (value: unknown) => asserts value is AsyncGenerator;
 	nativePromise: <T = unknown>(value: unknown) => asserts value is Promise<T>;
 	promise: <T = unknown>(value: unknown) => asserts value is Promise<T>;
 	generatorFunction: (value: unknown) => asserts value is GeneratorFunction;
+	asyncGeneratorFunction: (value: unknown) => asserts value is AsyncGeneratorFunction;
 	// eslint-disable-next-line @typescript-eslint/ban-types
 	asyncFunction: (value: unknown) => asserts value is Function;
 	// eslint-disable-next-line @typescript-eslint/ban-types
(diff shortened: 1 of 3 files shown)


test/test.ts
```diff
@@ -572,10 +591,33 @@ test('is.generator', t => {
 	testType(t, 'generator');
 });
 
+test('is.asyncGenerator', t => {
+	testType(t, 'asyncGenerator');
+
+	const fixture = (async function * () {
```
gioragutt: what's the added value of these? here you're not testing whether this is the type or not, but whether the implementation of the type "works as promised".

I think it's about the same as checking if a `string` has a `length` property...

test/test.ts
```diff
@@ -572,10 +591,33 @@ test('is.generator', t => {
...
+});
+
 test('is.generatorFunction', t => {
 	testType(t, 'generatorFunction', ['function']);
 });
 
+test('is.asyncGeneratorFunction', t => {
+	testType(t, 'asyncGeneratorFunction', ['function']);
+
+	const fixture = async function * () {
```
gioragutt: same

test/test.ts
```diff
@@ -189,6 +190,15 @@ const types = new Map<string, Test>([
 		],
 		typename: TypeName.Generator
 	}],
+	['asyncGenerator', {
+		is: is.asyncGenerator,
+		fixtures: [
+			(async function * () {
```
gioragutt: are you sure `function * ()` is the correct formatting here? with the spaces before **and** after the `*`?

if this is the formatting enforces by the tooling, ignore this comment

source/index.ts
```diff
@@ -153,6 +157,8 @@ is.promise = (value: unknown): value is Promise<unknown> => is.nativePromise(val
 
 is.generatorFunction = isObjectOfType<GeneratorFunction>(TypeName.GeneratorFunction);
 
+is.asyncGeneratorFunction = (value: unknown): value is ((...args: any[]) => Promise<unknown>) => getObjectType(value) === TypeName.AsyncGeneratorFunction;
```
gioragutt:
```typescript
((...args: any[]) => Promise<unknown>)
```
* is this the correct return type?
* such as there is a `GeneratorFunction` type (like above, L158), isn't there an `AsyncGeneratorFunction`?

test/test.ts
forresst: Yes, it's XO! He requests that the format be like this

test/test.ts
forresst: I did the same test as for asyncFunction (old L564, new L583). It's to keep the same logic.

test/test.ts
forresst: I did the same test as for asyncFunction (old L564, new L583). It's to keep the same logic.

source/index.ts
forresst: Unless I'm mistaken, it's an async function, it must return a promise like asyncFunction (old L156, New L162)

forresst:
@gioragutt I've left comments, I stay your tuned to discuss them.

sindresorhus:
@forresst Note https://github.com/sindresorhus/is/commit/0c3f11038638920377c44b0832ad0a89088bfe16. Could you also add this to the assertion stuff?

forresst:
@sindresorhus It's done, I've added the assertion

sindresorhus:
Perfect. Thank you 🙌

PR #101, opened 2020-01-21
Upgrade dependencies

joelpurra:
Attempts to fix linter parsing issues stemming from changes in typescript v3.7.

- Upgrade `eslint`-related packages.
- Upgrade `typescript` and related `tsconfig`.
- Fixes linting errors due to upgraded rules.
- Note that you might need to remove `node_modules` and run `npm install` again to properly clear linting caches.
- Include the `test` directory in the typescript project to fix the below parsing error.

> Parsing error: `"parserOptions.project"` has been set for `@typescript-eslint/parser`.
> The file does not match your project config: `test/test.ts`.
> The file must be included in at least one of the projects provided.

See

- https://github.com/typescript-eslint/typescript-eslint/pull/1045
- https://github.com/microsoft/TypeScript/commits/v3.7.5
- https://github.com/sindresorhus/tsconfig/issues/7
- https://github.com/xojs/eslint-config-xo-typescript/issues/20

### DIFF
--- a/tsconfig.xo.json
+++ b/tsconfig.xo.json
@@ -1,0 +1,6 @@
+{
+	"extends": "./tsconfig.json",
+	"include": [
+		"test"
+	]
+}
--- a/package.json
+++ b/package.json
@@ -46,20 +46,20 @@
 		"types"
 	],
 	"devDependencies": {
-		"@sindresorhus/tsconfig": "^0.6.0",
+		"@sindresorhus/tsconfig": "^0.7.0",
 		"@types/jsdom": "^12.2.4",
 		"@types/node": "^12.12.6",
 		"@types/zen-observable": "^0.8.0",
-		"@typescript-eslint/eslint-plugin": "^1.11.0",
-		"@typescript-eslint/parser": "^1.11.0",
+		"@typescript-eslint/eslint-plugin": "^2.17.0",
+		"@typescript-eslint/parser": "^2.17.0",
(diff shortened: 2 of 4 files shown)


tsconfig.json
```diff
@@ -9,6 +9,7 @@
 		]
 	},
 	"include": [
-		"source"
+		"source",
+		"test"
```
sindresorhus: Adding another directory here means you need to update https://github.com/sindresorhus/is/blob/aeb3f74d653cf745abf2460f0896207de00c6059/package.json#L12 and https://github.com/sindresorhus/is/blob/aeb3f74d653cf745abf2460f0896207de00c6059/package.json#L22, and maybe more. As it will no longer compile `source` to the top-level of `dist`. It would be nice to keep `source` compiling to `dist` directly.

joelpurra: Quite right -- only ran the local tests and didn't notice this change. Attempting to use the workaround from https://github.com/xojs/eslint-config-xo-typescript/issues/15; force-pushing an update.

tsconfig.xo.json
```diff
@@ -0,0 +1,6 @@
+{
+	"extends": "./tsconfig.json",
+	"include": [
+		"test"
```
sindresorhus: Shouldn't this also have `source`?

sindresorhus:
Thanks for doing this 👌

Issue #102, opened 2020-01-22
Add generic type parameters


joelpurra:
> What do you think about adding generic type parameters to some of the `is` methods too?

_Originally posted by @sindresorhus in https://github.com/sindresorhus/is/pull/97_

joelpurra:
Would be nice to propagate some more generic typings for the cases when typescript already knows about them, inferred/implicitly or provided/explicitly.

Note that `is` (and now the associated assertions) "actively" perform type checking work _at runtime_. The current implicit default is that generic types resolve to `unknown`<sup>1</sup>. This is a sane default, which is meant to force the user to perform further type checks before using the value. Allowing developers to define a generic type (inferred or explicit) _at compile-time_ can _hide_ the need to perform runtime type checks, leading to unexpected behavior. It is therefore up to developer to be aware and use explicit generics responsibly.

<sup>1</sup> Apart from a few assertion methods where I jumped the gun in #97.

## Examples

(Using a patched `is` with propagated generic types for both `is` and assertions.)

Works as expected when not providing an explicit `T`.

```typescript
(async function () {
    const t = Promise.resolve("an unexpected string value");
    // t is inferred at compile-time to be of type Promise<string>
    assert.promise(t);
    // t is (still) inferred at compile-time to be of type Promise<string>
    // t is known at runtime to be a Promise resolving to an unknown type

    const r = await t;
    // r is inferred at compile-time to be of type string
    console.log(r * 2);
    // does not compile as r is assumed to be of type unknown
}());
```

Showing intersection type mixup leading to unexpected behavior.

```typescript
(async function () {
    const t = Promise.resolve("an unexpected string value");
    // t is inferred at compile-time to be of type Promise<string>
    assert.promise<number>(t);
    // t is inferred at compile-time to be of intersection type Promise<string> & Promise<number>

    const r = await t;
    // r is assumed to be of type never, which is the intersection between string and number
    console.log(r * 2);
    // still (unexpectedly?) compiles, but logs NaN
}());
```

Showing that generic types are not checked at runtime.

```typescript
(async function () {
    const t: unknown = Promise.resolve("an unexpected string value");
    // t is inferred at compile-time to be of type unknown
    assert.promise<number>(t);
    // t is known at runtime to be a Promise, but is incorrectly assumed to be Promise<number> resolving to a number

    const r = await t;
    // r is assumed to be of type number
    console.log(r * 2);
    // logs NaN
}());
```

[...]

PR #103, opened 2020-01-22
Propagate generic types where possible

joelpurra:
- Replaces the hardcoded `unknown` type in for example `is.promise(value)` with `is.promise<T>(value)` so that a known (inferred or explicit) generic type `T` can be propagated in the type system.
  - The same goes for `Key` and `Value` types in for example `Map<Key, Value>`, and so on.
  - Also replacing `{[key: string]: unknown}` and similar with
`Record<Key, Value>` as appropriate.
  - Kept `is.plainObject` as `Record<string, Value>`, although the
`Key` can also be a `symbol`.
- If no inferred/explicit generic type is provided, the (current) default/implicit generic type is preserved. It is usually `unknown`.
- Use with caution, as `is` (nor the associated assertions) can not check generic type validity at runtime. It is a compile-time check performed by typescript, based on the developer-provided types (inferred or explicit).
- Making the wrong compile-time assumption about `T` can result in value with types which differ at runtime, leading to unexpected behavior. It is safer to stick with `unknown` generic types until they have been verified at runtime. This is particularly important when it comes to verifying the types of arbitrary inputs from users and external systems.
- No tests added, as changes should be compile-time only.

See

- https://www.typescriptlang.org/docs/handbook/generics.html
- https://github.com/sindresorhus/is/issues/102

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -135,25 +135,25 @@ is.buffer = (value: unknown): value is Buffer => !is.nullOrUndefined(value) && !
 
 is.nullOrUndefined = (value: unknown): value is null | undefined => is.null_(value) || is.undefined(value);
 is.object = (value: unknown): value is object => !is.null_(value) && (typeof value === 'object' || is.function_(value));
-is.iterable = (value: unknown): value is IterableIterator<unknown> => !is.nullOrUndefined(value) && is.function_((value as IterableIterator<unknown>)[Symbol.iterator]);
+is.iterable = <T = unknown>(value: unknown): value is IterableIterator<T> => !is.nullOrUndefined(value) && is.function_((value as IterableIterator<T>)[Symbol.iterator]);
 
-is.asyncIterable = (value: unknown): value is AsyncIterableIterator<unknown> => !is.nullOrUndefined(value) && is.function_((value as AsyncIterableIterator<unknown>)[Symbol.asyncIterator]);
+is.asyncIterable = <T = unknown>(value: unknown): value is AsyncIterableIterator<T> => !is.nullOrUndefined(value) && is.function_((value as AsyncIterableIterator<T>)[Symbol.asyncIterator]);
 
 is.generator = (value: unknown): value is Generator => is.iterable(value) && is.function_(value.next) && is.function_(value.throw);
 
-is.nativePromise = (value: unknown): value is Promise<unknown> =>
-	isObjectOfType<Promise<unknown>>(TypeName.Promise)(value);
+is.nativePromise = <T = unknown>(value: unknown): value is Promise<T> =>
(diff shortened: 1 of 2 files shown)


source/index.ts
```diff
@@ -590,12 +590,12 @@ export const assert: Assert = {
 	emptyString: (value: unknown): asserts value is '' => assertType(is.emptyString(value), AssertionTypeDescription.emptyString, value),
 	nonEmptyString: (value: unknown): asserts value is string => assertType(is.nonEmptyString(value), AssertionTypeDescription.nonEmptyString, value),
 	emptyStringOrWhitespace: (value: unknown): asserts value is string => assertType(is.emptyStringOrWhitespace(value), AssertionTypeDescription.emptyStringOrWhitespace, value),
-	emptyObject: (value: unknown): asserts value is {[key: string]: never} => assertType(is.emptyObject(value), AssertionTypeDescription.emptyObject, value),
-	nonEmptyObject: (value: unknown): asserts value is {[key: string]: unknown} => assertType(is.nonEmptyObject(value), AssertionTypeDescription.nonEmptyObject, value),
+	emptyObject: <Key extends keyof any = string>(value: unknown): asserts value is Record<Key, never> => assertType(is.emptyObject(value), AssertionTypeDescription.emptyObject, value),
```
sindresorhus: Why `any` and not `unknown`?

sindresorhus:
> Use with caution, as is (nor the associated assertions) can not check generic type validity at runtime. It is a compile-time check performed by typescript, based on the developer-provided types (inferred or implicit).

We should definitely have a warning in the readme about this.

source/index.ts
joelpurra: To match `Record<,>`. Also, `unknown` wouldn't work here as default types would not work. In this case it's reported as `Type 'string' does not satisfy the constraint 'never'.`

https://github.com/microsoft/TypeScript/blob/v3.7.5/lib/lib.es5.d.ts#L1466-L1471

brandon93s:
👍 from me.  Thank you for the examples in https://github.com/sindresorhus/is/issues/102.  Please add the above noted warning to the readme.

gioragutt:
@sindresorhus regarding the warning, it's true for anything that doesn't really from from your code, like `JSON.parse`ing a string... It's not really an `is` thing, but a typescript thing.

I do agree it's important, just in case.

source/index.ts
sindresorhus: 👍

joelpurra:
- Added a section about propagating generic type parameters in the readme.
- Added a link at the top of the readme, including a brief warning.
- Included an example based on wrongly assuming/correctly asserting a resolved `Promise<T>`. It might not be the best example; am open to suggestions.

PR #110, opened 2020-04-13
Fix #109: `is.numericString` behaves correctly on whitespaces

### LINKED ISSUE #109: numericString gives true for string with spaces
This code  --
```js
is.numericString(' ');
is.numericString('     ');
```
returns true for both cases. 

I was expecting them to return false. Right now, the check is this --
```js
is.string(value) && value.length > 0 && !Number.isNaN(Number(value));
```
`Number` coerces whitespaces to 0 which is not taken into account right now.

I can make a PR if this was not the original intended behavior.

ArfatSalman:
I have added the required test cases. @gioragutt Let me know if I you want me to add more test cases.

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -130,7 +130,7 @@ is.boolean = (value: unknown): value is boolean => value === true || value === f
 is.symbol = isOfType<symbol>('symbol');
 
 is.numericString = (value: unknown): value is string =>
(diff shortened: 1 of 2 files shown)


ArfatSalman:
@papb 

> I think you should use \t instead of a literal tab to make it much more readable. You could even omit the comments in this case

Done.

PR #117, opened 2020-06-20
Make the `is()` function type-safe

bjornstar:
This is the functional change that was removed from #113 

There was an unsafe cast in getObjectType which could allow any string to be returned. This was used as an output in the primary `is` function.

The primary function `is` now always returns a `TypeName`.

The other breaking change here is the return value of `HTMLElement` for dom elements like `div` or `span`. Previously it would return `HTMLDivElement` or `HTMLSpanElement`, but now they all return `HTMLElement`.

### DIFF
--- a/test/test.ts
+++ b/test/test.ts
@@ -1284,10 +1284,10 @@ test('is.domElement', t => {
 		'script'
 	];
 
-	tagNames.forEach(tagName => {
+	for (const tagName of tagNames) {
 		const domElement = createDomElement(tagName);
 		t.is(is(domElement), 'HTMLElement');
-	});
+	}
 });
 
 test('is.observable', t => {
--- a/test/test.ts
+++ b/test/test.ts
@@ -1275,19 +1275,19 @@ test('is.domElement', t => {
 		assert.domElement({nodeType: 1, nodeName: 'div'});
 	});
 
-	const htmlTagNameToTypeName = {
-		div: 'HTMLDivElement',
-		input: 'HTMLInputElement',
-		span: 'HTMLSpanElement',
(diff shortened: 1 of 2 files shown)


test/test.ts
```diff
@@ -1275,19 +1275,19 @@ test('is.domElement', t => {
...
+		'div',
+		'input',
+		'span',
+		'img',
+		'canvas',
+		'script'
+	];
 
-	for (const [tagName, typeName] of Object.entries(htmlTagNameToTypeName)) {
+	tagNames.forEach(tagName => {
```
sindresorhus, line 1287: Use a `for-of` loop.

sindresorhus:
> The other breaking change here is the return value of HTMLElement for dom elements like div or span. Previously it would return HTMLDivElement or HTMLSpanElement, but now they all return HTMLElement.

Can you elaborate on *why* we should change this? What's the benefit of just returning `HTMLElement`?

bjornstar:
> > The other breaking change here is the return value of HTMLElement for dom elements like div or span. Previously it would return HTMLDivElement or HTMLSpanElement, but now they all return HTMLElement.
> 
> Can you elaborate on _why_ we should change this? What's the benefit of just returning `HTMLElement`?

The current type guard is written for the `HTMLElement` type so it's consistent with what is implemented.

To be considered type safe for [all 119 HTML tags](https://github.com/microsoft/TypeScript/blob/v3.9.5/lib/lib.dom.d.ts#L19224) would require a significant increase to the size and scope of the project.

sindresorhus:
Thanks :)

Released: https://github.com/sindresorhus/is/releases/tag/v3.0.0

PR #119, opened 2020-07-17
Add `is.array` overload that supports asserting array items

### LINKED ISSUE #112: Is it possible to assert that a variable is an array of strings?
Is it possible to assert that a variable is an array of strings, ideally with automatic typescript type assertion to `string[]`? I could not find information about this in the readme. It does not seem possible. 

If yes, this issue is a readme clarification request. If not, this issue is a feature request :grimacing:

Arnovsky:
Fixes #112

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -98,8 +98,8 @@ Note: `is.number(NaN)` returns `false`. This intentionally deviates from `typeof
 Returns true if `value` is an array and all of its items match the assertion (if provided).
 
 ```js
-is.array(value) // validate `value` is an array
-is.array(value, is.number) // validate `value` is an array and all of it's items are numbers.
+is.array(value); // Validate `value` is an array.
+is.array(value, is.number); // Validate `value` is an array and all of its items are numbers.
 ```
 
 ##### .function(value)
--- a/readme.md
+++ b/readme.md
@@ -99,7 +99,7 @@ Returns true if `value` is an array and all of it's items match the assertion.
 
 ```js
 is.array(value) // validate `value` is an array
-is.array(value, is.number) // validate that value is an array and all of it's items are numbers.
+is.array(value, is.number) // validate `value` is an array and all of it's items are numbers.
 ```
 
 ##### .function(value)
--- a/readme.md
+++ b/readme.md
@@ -93,7 +93,15 @@ Note: `is.number(NaN)` returns `false`. This intentionally deviates from `typeof
 
 #### Built-in types
 
-##### .array(value)
+##### .array(value, assertion)
+
+Returns true if `value` is an array and all of it's items match the assertion.
+
+```js
+is.array(value) // validate `value` is an array
+is.array(value, is.number) // validate that value is an array and all of it's items are numbers.
+```
+
 ##### .function(value)
 ##### .buffer(value)
 ##### .object(value)
--- a/source/index.ts
+++ b/source/index.ts
@@ -185,7 +185,16 @@ is.symbol = isOfType<symbol>('symbol');
 is.numericString = (value: unknown): value is string =>
 	is.string(value) && !is.emptyStringOrWhitespace(value) && !Number.isNaN(Number(value));
 
-is.array = Array.isArray;
+is.array = <T = unknown>(value: any, assertion?: (value: T) => void): value is T[] =>  {
+	if (!Array.isArray(value)) {
+		return false;
+	}
+	if (!assertion) {
+		return true;
+	}
+	return value.every(assertion);
+};
+
 is.buffer = (value: unknown): value is Buffer => (value as any)?.constructor?.isBuffer?.(value) ?? false;
 
 is.nullOrUndefined = (value: unknown): value is null | undefined => is.null_(value) || is.undefined(value);
@@ -463,7 +472,7 @@ interface Assert {
 	boolean: (value: unknown) => asserts value is boolean;
 	symbol: (value: unknown) => asserts value is symbol;
 	numericString: (value: unknown) => asserts value is string;
-	array: <T = unknown>(value: unknown) => asserts value is T[];
+	array: <T = unknown>(value: unknown, assertion?: (element: unknown) => void) => asserts value is T[];
 	buffer: (value: unknown) => asserts value is Buffer;
 	nullOrUndefined: (value: unknown) => asserts value is null | undefined;
 	object: <Key extends keyof any = string, Value = unknown>(value: unknown) => asserts value is Record<Key, Value>;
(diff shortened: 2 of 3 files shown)


papb:
I see that your code uses `assert.array([1, 2], is.number)` instead of `assert.array([1, 2], assert.number)`, why?

Arnovsky:
> 
> 
> I see that your code uses `assert.array([1, 2], is.number)` instead of `assert.array([1, 2], assert.number)`, why?

assert.array is based on is.array which receives a predicate (wrongly?) named `assertion`.
is this behavior OK? Since we need to support `is.array` as it returns a boolean and not throws, I didn't see a better API without major changes.

gioragutt:
I'd say that `is.array` should receive a type guard, whereas `assert.array` should receive an... assert guard? I don't remember the term.

Although, logically, `assert.array` should work with both predicates and `assert` methods, it kinda makes sense...

source/index.ts
```diff
@@ -185,7 +185,18 @@ is.symbol = isOfType<symbol>('symbol');
 is.numericString = (value: unknown): value is string =>
 	is.string(value) && !is.emptyStringOrWhitespace(value) && !Number.isNaN(Number(value));
 
-is.array = Array.isArray;
+is.array = <T = unknown>(value: unknown, assertion?: (value: T) => void): value is T[] => {
```
gioragutt, line 188:
```suggestion
is.array = <T = unknown>(value: unknown, assertion?: (value: T) => value is T): value is T[] => {
```

On the `is` method, we're not asserting anything, and we don't throw exceptions.
Your `assertion` parameter returns `void`, which is not what is expected from `is.array`, and you can actually give the wrong time of method as the `assertion` param and it would allow it (and might return the wrong value).

The `assert.array` methods should in fact work with `asserts ...` in the typing.

source/index.ts
```diff
@@ -553,7 +564,7 @@ export const assert: Assert = {
 	boolean: (value: unknown): asserts value is boolean => assertType(is.boolean(value), 'boolean', value),
 	symbol: (value: unknown): asserts value is symbol => assertType(is.symbol(value), 'symbol', value),
 	numericString: (value: unknown): asserts value is string => assertType(is.numericString(value), AssertionTypeDescription.numericString, value),
-	array: <T = unknown>(value: unknown): asserts value is T[] => assertType(is.array(value), 'Array', value),
+	array: <T = unknown>(value: unknown, assertion?: (element: unknown) => asserts element is T): asserts value is T[] => assertType(is.array(value, assertion), 'Array', value),
```
gioragutt, line 567: I don't think that `is.array` and `assert.array` are compatible.

We have two options:

1. The `assertion` parameter should be a `type guard` for both `is.array` and `assert.array`, and it would work.
2. You'd keep the `assertion` parameter as an `asserts ...` and the `assert.array` implementation would be less simple than the rest of the methods.

gioragutt: Also, as we talked, a better API would be to use `type guards` in `is.array`, since we don't want to `try-catch` the assertion, as we might end up ignoring actual exceptions thrown by user code.

source/index.ts
```diff
@@ -553,7 +564,14 @@ export const assert: Assert = {
 	boolean: (value: unknown): asserts value is boolean => assertType(is.boolean(value), 'boolean', value),
 	symbol: (value: unknown): asserts value is symbol => assertType(is.symbol(value), 'symbol', value),
 	numericString: (value: unknown): asserts value is string => assertType(is.numericString(value), AssertionTypeDescription.numericString, value),
-	array: <T = unknown>(value: unknown): asserts value is T[] => assertType(is.array(value), 'Array', value),
+	array: <T = unknown>(value: unknown, assertion?: (element: unknown) => asserts element is T): asserts value is T[] => {
+		const assert: (condition: boolean, description: string, value: unknown) => asserts condition = assertType;
```
Arnovsky, line 568: I did this for prevent the `Assertions require every name in the call target to be declared with an explicit type annotation.ts(2775)` error.

If there is something I am missing here (which I certainly think I am) let me know!

sindresorhus:
This needs to be documented in the readme.

readme.md
```diff
@@ -93,7 +93,15 @@ Note: `is.number(NaN)` returns `false`. This intentionally deviates from `typeof
 
 #### Built-in types
 
-##### .array(value)
+##### .array(value, assertion)
+
+Returns true if `value` is an array and all of it's items match the assertion.
```
papb, line 98:
```suggestion
##### .array(value, assertion?)

Returns true if `value` is an array and all of its items match the assertion (if provided).
```

papb:
I would also like to request some tests to ensure that the type assertion works in TS. For example, the following code must compile in TS:

```ts
function foo(arg: unknown) {
  assert.array(arg, is.number);
  console.log(arg.toFixed(2));
}
```

Arnovsky:
> 
> 
> I would also like to request some tests to ensure that the type assertion works in TS. For example, the following code must compile in TS:
> 
> ```ts
> function foo(arg: unknown) {
>   assert.array(arg, is.number);
>   console.log(arg[0].toFixed(2));
> }
> ```
> 
> Just like the following already works:
> 
> ```ts
> function foo(arg: unknown) {
>   assert.number(arg);
>   console.log(arg.toFixed(2));
> }
> ```

As @gioragutt mentioned earlier the API will be `assert.array(arg, assert.number);`, As for the testing, what about something like this:

```ts
t.notThrows(() => {
	const x: any[] = [1, 2, 3];
	assert.array(x, assert.number);
	x[0].toFixed(0);
});

t.notThrows(() => {
	const x: any[] = [1, 2, 3];
	if (is.array<number>(x, is.number)) {
		x[0].toFixed(0);
	}
});
```
We don't actually need `notThrows`, only need to see if this code compiles. Which IMO does what we want here.

source/index.ts
```diff
@@ -185,7 +185,18 @@ is.symbol = isOfType<symbol>('symbol');
...
+is.array = <T = unknown>(value: unknown, assertion?: (value: T) => value is T): value is T[] => {
+	if (!Array.isArray(value)) {
+		return false;
+	}
+
+	if (!assertion) {
+		return true;
+	}
+
+	return value.every(assertion);
```
bjornstar, line 197:
```suggestion
	return Array.isArray(value) && !assertion || value.every(assertion);
```

Arnovsky: Hi, I know that this can be shortened to a single line, but IMO this feels less readable.

papb:
@Arnovsky Nice, however I think `unknown` should be used instead of `any`, since the following would already compile anyway:

```ts
const x: any[] = [1, 2, 3];
x[0].toFixed(0);
```

source/index.ts
papb: I agree with @Arnovsky here... I had to triple-check to be convinced this was equivalent. Also the [no-mixed-operators](https://eslint.org/docs/rules/no-mixed-operators) rule would complain here by the way. I vote for keeping it as is :)

Arnovsky:
@papb LGTY?

gioragutt:
@sindresorhus we good to merge?

papb:
> @papb LGTY?

@Arnovsky Yes, awesome!! Sorry for not replying earlier!

Thank you all very much! It is working perfectly.

PR #120, opened 2020-07-19
Extract necessary functions from `type-fest`

### LINKED ISSUE #85: Refactor to use type-fest
Please see https://github.com/sindresorhus/type-fest/issues/10

### LINKED ISSUE #108: Property 'observable' does not exist on type 'SymbolConstructor'
> Since I found a solution to my problem, this issue is rather informational.

On a small **mono-repo** (`lerna@3.20.2`) project:

- node.js v12.16.1 and npm 6.14.2
- `typescript@^3.8.3` as root `devDependencies`
- `@types/node@^12.12.29` as **root** `devDependencies`
- `@sindre/sorhus@^2.1.0` as **package** `dependencies`

```shell
❯ tsc --build tsconfig.json
node_modules/@sindresorhus/is/dist/index.d.ts:135:13 - error TS2339: Property 'observable' does not exist on type 'SymbolConstructor'.

135     [Symbol.observable](): ObservableLike;
                ~~~~~~~~~~
```

From what I could gather, `observable` on type `SymbolConstructor` is defined in `@types/node@12.12.29` (which is installed in the **root** `node_modules`), but not in `@types/node@13.7.7` (which ends up being installed in the **package's** `node_module` as a side effect.)

One kludgy quick-fix could be to add `@types/node@^12.12.29` in the package's `devDependencies`

Another, arguably cleaner solution is to use `lerna bootstrap --hoist` so that **root** node_modules prevails.

Richienb:
Fixes #85, Fixes #108

// @sindresorhus

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -2,7 +2,7 @@
 /// <reference lib="dom"/>
 /// <reference types="node"/>
 
-export type Class<T = unknown> = new (...args: any[]) => T;
+import {Class, TypedArray, ObservableLike, Primitive} from 'type-fest';
 
 const typedArrayTypeNames = [
 	'Int8Array',
@@ -17,19 +17,6 @@ const typedArrayTypeNames = [
 	'BigInt64Array',
 	'BigUint64Array'
 ] as const;
-
-export type TypedArray =
-	| Int8Array
-	| Uint8Array
-	| Uint8ClampedArray
-	| Int16Array
-	| Uint16Array
-	| Int32Array
-	| Uint32Array
-	| Float32Array
-	| Float64Array
-	| BigInt64Array
-	| BigUint64Array;
 
 type TypedArrayTypeName = typeof typedArrayTypeNames[number];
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
I have changed my mind about depending on `type-fest`. I don't think it's worth adding that dependency just for 3 simple types. Can you instead create a types.ts file and copy-paste the types we need from `type-fest` there?

Richienb:
@sindresorhus Fixed!

source/index.ts
```diff
@@ -2,7 +2,7 @@
 /// <reference lib="dom"/>
 /// <reference types="node"/>
 
-export type Class<T = unknown> = new (...args: any[]) => T;
+import {Class, TypedArray, ObservableLike, Primitive} from './types';
```
sindresorhus, line 5: You need to export these again as they were previously exported to end-users.

Richienb: @sindresorhus Fixed!

PR #122, opened 2020-08-16
Improve import for VS Code

iyegoroff:
When 'main' & 'types' are pointing to 'dist' folder vscode will autoimport the package like this:
```ts
import is from '@sindresorhus/is/dist'
```


gioragutt:
So what's the behavior today? With `main` and `types` pointing to `/dist`?

sindresorhus:
That's really a VS Code bug, but I'm fine with being explicit about the paths.

PR #126, opened 2020-10-06
Improve `is.plainObject` TypeScript type

RocktimSaikia:
Talked about in https://github.com/sindresorhus/is-plain-obj/issues/9#issuecomment-693722367.

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -260,7 +260,8 @@ is.primitive = (value: unknown): value is Primitive => is.null_(value) || isPrim
 is.integer = (value: unknown): value is number => Number.isInteger(value as number);
 is.safeInteger = (value: unknown): value is number => Number.isSafeInteger(value as number);
 
-is.plainObject = <Value = unknown, Key extends keyof any = string | number | symbol>(value: unknown): value is Record<Key, Value> => {
+type ObjectKey = string | number | symbol;
+is.plainObject = <Value = unknown>(value: unknown): value is Record<ObjectKey, Value> => {
 	// From: https://github.com/sindresorhus/is-plain-obj/blob/master/index.js
 	if (toString.call(value) !== '[object Object]') {
 		return false;
@@ -494,7 +495,7 @@ interface Assert {
 	primitive: (value: unknown) => asserts value is Primitive;
 	integer: (value: unknown) => asserts value is number;
 	safeInteger: (value: unknown) => asserts value is number;
-	plainObject: <Value = unknown, Key extends keyof any = string | number | symbol>(value: unknown) => asserts value is Record<Key, Value>;
+	plainObject: <Value = unknown>(value: unknown) => asserts value is Record<ObjectKey, Value>;
 	typedArray: (value: unknown) => asserts value is TypedArray;
 	arrayLike: <T = unknown>(value: unknown) => asserts value is ArrayLike<T>;
 	domElement: (value: unknown) => asserts value is HTMLElement;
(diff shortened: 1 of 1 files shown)


source/index.ts
```diff
@@ -260,7 +260,7 @@ is.primitive = (value: unknown): value is Primitive => is.null_(value) || isPrim
 is.integer = (value: unknown): value is number => Number.isInteger(value as number);
 is.safeInteger = (value: unknown): value is number => Number.isSafeInteger(value as number);
 
-is.plainObject = <Value = unknown>(value: unknown): value is Record<string, Value> => {
+is.plainObject = <Value = unknown>(value: unknown): value is Record<string | number | symbol, Value> => {
```
gioragutt, line 263:
```suggestion
is.plainObject = <Value = unknown, Key = string | number | symbol>(value: unknown): value is Record<Key, Value> => {
```

Just for cases people know that they might have just strings, and it wouldn't mess with their types. Default behavior would change (as this is a breaking change, sort of), but people would be able to easily decide the key type.

source/index.ts
```diff
@@ -494,7 +494,7 @@ interface Assert {
 	primitive: (value: unknown) => asserts value is Primitive;
 	integer: (value: unknown) => asserts value is number;
 	safeInteger: (value: unknown) => asserts value is number;
-	plainObject: <Value = unknown>(value: unknown) => asserts value is Record<string, Value>;
+	plainObject: <Value = unknown>(value: unknown) => asserts value is Record<string | number | symbol, Value>;
```
gioragutt, line 497: Same as the first comment

gioragutt (changes requested):

source/index.ts
papb: Interesting, is there any difference between `keyof any` and `string | number | symbol`?

source/index.ts
```diff
@@ -260,7 +260,7 @@ is.primitive = (value: unknown): value is Primitive => is.null_(value) || isPrim
 is.integer = (value: unknown): value is number => Number.isInteger(value as number);
 is.safeInteger = (value: unknown): value is number => Number.isSafeInteger(value as number);
 
-is.plainObject = <Value = unknown>(value: unknown): value is Record<string | number | symbol, Value> => {
+is.plainObject = <Value = unknown, Key extends keyof any = string | number | symbol>(value: unknown): value is Record<Key, Value> => {
```
gioragutt, line 263: 1. These changes should be tested, to see the the typing (default vs non-default) work as expected
2. The `assert` method should change as well

gioragutt (changes requested):

source/index.ts
RocktimSaikia: just changed and pushed

source/index.ts
sindresorhus: > Interesting, is there any difference between keyof any and string | number | symbol?

@gioragutt ^

sindresorhus: Should we maybe move `string | number | symbol` to a `type` called `ObjectKey` since it's used in multiple places?

gioragutt: AFAIR `keyof any === string |number | symbol`

sindresorhus: Then I would prefer `string | number | symbol`.

RocktimSaikia: So we should stick with the original commit where it was `string | number | symbol` ?

sindresorhus: > Should we maybe move string | number | symbol to a type called ObjectKey since it's used in multiple places?

I need opinions on this.

RocktimSaikia: Yeah, that would be pretty handy in my opinion.

sindresorhus: Let's do that then.

RocktimSaikia: @sindresorhus  The suggested `ObjectKey` type is added.

PR #132, opened 2021-04-11
Fix assertion message for `.all` and `.any`

### LINKED ISSUE #131: Falsy values are treated as `Array` type
```js
const is = require('@sindresorhus/is');

is.assert.any([is.urlInstance], undefined);
```

Actual:

```
TypeError: Expected value which is `predicate returns truthy for any value`, received value of type `Array`.
```

Expected:

```
TypeError: Expected value which is `predicate returns truthy for any value`, received value of type `undefined`.
```

scraggo:
This PR fixes #131 

Approach: Add an options object to `assertType`. This allows a custom failure message from the calling code. It's a function that returns a string and passed in as `getValuesMessage` to the options object.

What do you think of the approach?

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -388,10 +388,17 @@ is.any = (predicate: Predicate | Predicate[], ...values: unknown[]): boolean => 
 
 is.all = (predicate: Predicate, ...values: unknown[]): boolean => predicateOnArray(Array.prototype.every, predicate, values);
 
-const assertType = (condition: boolean, description: string, value: unknown, options: { getValuesMessage?: () => string } = {}): asserts condition => {
+const assertType = (condition: boolean, description: string, value: unknown, options: { multipleValues?: boolean } = {}): asserts condition => {
 	if (!condition) {
-		const {getValuesMessage} = options;
-		const valuesMessage = getValuesMessage ? getValuesMessage() : `received value of type \`${is(value)}\``;
+		const {multipleValues} = options;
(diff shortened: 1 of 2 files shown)


source/index.ts
```diff
@@ -388,9 +388,11 @@ is.any = (predicate: Predicate | Predicate[], ...values: unknown[]): boolean =>
 
 is.all = (predicate: Predicate, ...values: unknown[]): boolean => predicateOnArray(Array.prototype.every, predicate, values);
 
-const assertType = (condition: boolean, description: string, value: unknown): asserts condition => {
+const assertType = (condition: boolean, description: string, value: unknown, options: { getValuesMessage?: () => string } = {}): asserts condition => {
```
sindresorhus, line 391: I don't see the point in nesting it in an options object. We're only using this internally. Can just be a normal argument.

sindresorhus (changes requested):

sindresorhus:
Does the issue also apply to `.all`? If so, we should fix it there too.

scraggo:
@sindresorhus I tried a slightly different approach in my latest push and applied it to `.all` as well.

I decided to keep the options object for this approach because it's now a boolean flag. `{multipleValues: true}` is more readable in my opinion than a bare boolean. This isn't a strong opinion and it's easily changed if you disagree 😄

sindresorhus:
Thanks :)

PR #138, opened 2021-09-07
Add `is.propertyKey`

### LINKED ISSUE #136: Proposal: is.propertyKey(value)
Just a handy type guard of the `PropertyKey` persuasion...

else I'll do with the lazy implementation:

```ts
const isPropertyKey = (value: unknown): value is PropertyKey => is.any([is.string, is.number, is.symbol], value);
```

PopGoesTheWza:
Fixes #136

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -430,7 +430,7 @@ Returns `true` if `value` is an odd integer.
 
 ##### .propertyKey(value)
 
-Returns `true` if `value` can be used as an object property key (either `string`, `number` or `symbol`.)
+Returns `true` if `value` can be used as an object property key (either `string`, `number`, or `symbol`.)
 
 ##### .any(predicate | predicate[], ...values)
 
--- a/readme.md
+++ b/readme.md
@@ -430,7 +430,7 @@ Returns `true` if `value` is an odd integer.
 
 ##### .propertyKey(value)
 
-Returns `true` if `value` can be used as an object property key (either `string`, `number`, or `symbol`.)
+Returns `true` if `value` can be used as an object property key (either `string`, `number`, or `symbol`).
 
 ##### .any(predicate | predicate[], ...values)
 
--- a/readme.md
+++ b/readme.md
@@ -428,6 +428,10 @@ Returns `true` if `value` is an even integer.
 
 Returns `true` if `value` is an odd integer.
 
+##### .propertyKey(value)
+
+Returns `true` if `value` can be used as an object property key (either `string`, `number` or `symbol`.)
+
 ##### .any(predicate | predicate[], ...values)
 
 Using a single `predicate` argument, returns `true` if **any** of the input `values` returns true in the `predicate`:
--- a/source/index.ts
+++ b/source/index.ts
@@ -360,9 +360,7 @@ is.nonEmptySet = <T = unknown>(value: unknown): value is Set<T> => is.set(value)
 is.emptyMap = (value: unknown): value is Map<never, never> => is.map(value) && value.size === 0;
 is.nonEmptyMap = <Key = unknown, Value = unknown>(value: unknown): value is Map<Key, Value> => is.map(value) && value.size > 0;
 
-/**
-`PropertyKey` is any value that can be used as an object key (string, number, or symbol)
-*/
+// `PropertyKey` is any value that can be used as an object key (string, number, or symbol)
 is.propertyKey = (value: unknown): value is PropertyKey => is.any([is.string, is.number, is.symbol], value);
 
 export type Predicate = (value: unknown) => boolean;
--- a/source/index.ts
+++ b/source/index.ts
@@ -359,7 +359,10 @@ is.nonEmptySet = <T = unknown>(value: unknown): value is Set<T> => is.set(value)
 
 is.emptyMap = (value: unknown): value is Map<never, never> => is.map(value) && value.size === 0;
 is.nonEmptyMap = <Key = unknown, Value = unknown>(value: unknown): value is Map<Key, Value> => is.map(value) && value.size > 0;
-/** Returns `true` if `value` can be used as an object property key (either a `string`, a `number` or a `symbol`.) */
+
+/**
+Returns `true` if `value` can be used as an object property key (either a `string`, a `number` or a `symbol`).
+*/
 is.propertyKey = (value: unknown): value is PropertyKey => is.any([is.string, is.number, is.symbol], value);
 
 export type Predicate = (value: unknown) => boolean;
--- a/source/index.ts
+++ b/source/index.ts
@@ -150,6 +150,8 @@ is.string = isOfType<string>('string');
 const isNumberType = isOfType<number>('number');
 is.number = (value: unknown): value is number => isNumberType(value) && !is.nan(value);
 
+is.propertyKey = (value: unknown): value is PropertyKey => is.any([is.string, is.number, is.symbol], value);
+
 is.bigint = isOfType<bigint>('bigint');
 
 // eslint-disable-next-line @typescript-eslint/ban-types
@@ -448,6 +450,7 @@ interface Assert {
 	undefined: (value: unknown) => asserts value is undefined;
 	string: (value: unknown) => asserts value is string;
 	number: (value: unknown) => asserts value is number;
+	propertyKey: (value: unknown) => asserts value is PropertyKey;
 	bigint: (value: unknown) => asserts value is bigint;
 	// eslint-disable-next-line @typescript-eslint/ban-types
 	function_: (value: unknown) => asserts value is Function;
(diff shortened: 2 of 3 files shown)


PopGoesTheWza:
@sindresorhus older version of Typescript may not include `Symbol` in the `PropertyKey` definition. Would that be an issue?

source/index.ts
```diff
@@ -150,6 +150,8 @@ is.string = isOfType<string>('string');
 const isNumberType = isOfType<number>('number');
 is.number = (value: unknown): value is number => isNumberType(value) && !is.nan(value);
 
+is.propertyKey = (value: unknown): value is PropertyKey => is.any([is.string, is.number, is.symbol], value);
```
sindresorhus, line 153: Kinda weird place to place it. This should come after all the native types.

source/index.ts
```diff
@@ -538,6 +541,8 @@ export const assert: Assert = {
 	undefined: (value: unknown): asserts value is undefined => assertType(is.undefined(value), 'undefined', value),
 	string: (value: unknown): asserts value is string => assertType(is.string(value), 'string', value),
 	number: (value: unknown): asserts value is number => assertType(is.number(value), 'number', value),
+	// ? Is an AssertionTypeDescription.propertyKey entry necessary?
```
sindresorhus, line 544: Why wouldn't it be?

test/test.ts
```diff
@@ -613,6 +613,13 @@ test('is.number', t => {
 	testType(t, 'number', ['integer', 'safeInteger', 'infinite']);
 });
 
+test('is.propertyKey', t => {
+	// ??? Unsure what is the correct syntax to `testType` multiple types. Also are the exclusion relevant?
```
sindresorhus, line 617: I recommend looking at the `testType` source and see if you can figure it out yourself first. I don't remember exactly how it works either.

source/index.ts
PopGoesTheWza: Would placing it after each `is.plainObject` occurences make you feel more confortable?

source/index.ts
PopGoesTheWza: Just askin'... You're the boss.

test/test.ts
PopGoesTheWza: As far as understand it, the test works and is 'good enough'

sindresorhus:
You need to add it to the readme.

source/index.ts
```diff
@@ -359,6 +359,7 @@ is.nonEmptySet = <T = unknown>(value: unknown): value is Set<T> => is.set(value)
 
 is.emptyMap = (value: unknown): value is Map<never, never> => is.map(value) && value.size === 0;
 is.nonEmptyMap = <Key = unknown, Value = unknown>(value: unknown): value is Map<Key, Value> => is.map(value) && value.size > 0;
+is.propertyKey = (value: unknown): value is PropertyKey => is.any([is.string, is.number, is.symbol], value);
```
sindresorhus, line 362: Needs a doc comment as it's not immediately obvious what it checks for.

sindresorhus (changes requested):

source/index.ts
```diff
@@ -360,6 +360,11 @@ is.nonEmptySet = <T = unknown>(value: unknown): value is Set<T> => is.set(value)
 is.emptyMap = (value: unknown): value is Map<never, never> => is.map(value) && value.size === 0;
 is.nonEmptyMap = <Key = unknown, Value = unknown>(value: unknown): value is Map<Key, Value> => is.map(value) && value.size > 0;
 
+/**
+Returns `true` if `value` can be used as an object property key (either a `string`, a `number` or a `symbol`).
```
sindresorhus, line 364: The text here should be exactly the same as in the readme. Notice the `a`'s.

gioragutt: @sindresorhus why is a comment needed in the cod? I don't see similar cases of comments in the code.

gioragutt, line 364: @sindresorhus I may have deleted your comment by mistake 🤭 sorry about that.

@PopGoesTheWza His comment was about the comment being like others.

IMO a more fitting comment would be to describe what a `PropertyKey` is, instead of the whole "returns true if...", that's good for the readme but unnecessary in the code IMO, since the reason for the comment here is that it's not clear what `PropertyKey` is at first glance.

test/test.ts
```diff
@@ -1490,6 +1490,12 @@ test('is.nonEmptyMap', t => {
 	});
 });
 
+test('is.propertyKey', t => {
+	testType(t, 'string', ['emptyString', 'numericString']);
+	testType(t, 'number', ['nan', 'integer', 'safeInteger', 'infinite']);
+	testType(t, 'symbol');
```
gioragutt, line 1496: @PopGoesTheWza

Looking at other tests, it seems that `testType` is to be used like the following:

```ts
test('is.XXX', t => {
	testType(t, 'XXX', [...exceptions]);
});
```

So given that, you should:

1. Add `propertyKey` to the `types` map (line 43 currently)
2. Update the test to the following:

```ts
test('is.propertyKey', t => {
	testType(t, 'propertyKey', [...exceptions]);
});
```

Where `exceptions` is probably:

```ts
[
	'string', 'emptyString', 'numericString', 
	'number', 'nan', 'integer', 'safeInteger', 'infinite', 
	'symbol',
]
```

Which is the merge of whatever is tests would also end up positive for your predicate.

readme.md
```diff
@@ -428,6 +428,10 @@ Returns `true` if `value` is an even integer.
 
 Returns `true` if `value` is an odd integer.
 
+##### .propertyKey(value)
+
+Returns `true` if `value` can be used as an object property key (either `string`, `number`, or `symbol`.)
```
gioragutt, line 433:
```suggestion
Returns `true` if `value` can be used as an object property key (either `string`, `number`, or `symbol`).
```

source/index.ts
PopGoesTheWza: Something like this?

> Checks if `value` safe to use as a property key (typically when working with `Object`, `Proxy` and `Reflect` methods)

gioragutt: Don't think it's needs to go that far, just to explain what `PropertyKey` is, something like:

> `PropertyKey` is any value that can be used as an object key (string, number, or symbol)

test/test.ts
PopGoesTheWza: @gioragutt it won't work. Assuming I create a new `Test`:

```ts
	['propertyKey', {
		is: is.propertyKey,
		assert: assert.propertyKey,
		fixtures: [
			undefined
		],
		typename: 'propertyKey'
	}],
```

1. Attribute `typename` needs to be a value from `type TypeName = ObjectTypeName | PrimitiveTypeName`. Neither of the two include `propertyKey` and adding it would bee semantically pointless (IMHO)
2. Attribute `typename` is used to filter entries from the `types` map (i.e. entries `number`, `nan`, `integer`, `safeInteger` and `infinite` all  share the same `typename`.) but it only support a single value. In this case, `typename` would need to support multiple values (i.e. `['string', 'number', 'symbol']`) in order to benefit from existing entries.
3. Modifying `testType()` in order to support many `typename`s is a bit complex and a task I would not venture without good reason

PopGoesTheWza:
> Never knew `PropertyKey` was a built-in type, always used `keyof any` instead.

Ouch! I used `string | number` before venturing in the Typescript standard libraries (which is full of good information). Support for `symbol` keys was added recently to Typescript.

test/test.ts
gioragutt: Oh, got it. Makes sense, guess I forgot the code. `testType` should be used to, well, test a *Type*.

Since `PropertyKey` isn't a type, but a union of types, `testType` does not fit.

However, it should still have tests that look something like:

```ts
test('is.propertyKey', t => {
	t.true(is.propertyKey('key'));
	t.true(is.propertyKey(5));
	t.true(is.propertyKey(Symbol('')));

	t.false(is.propertyKey(true));
	t.false(is.propertyKey({}));
	t.false(is.propertyKey([]));
	t.false(is.propertyKey(new Map()));
	t.false(is.propertyKey(new Set()));
});
```

If you see another case for either `true` cases or `false` cases feel free to add them.

source/index.ts
```diff
@@ -360,6 +360,11 @@ is.nonEmptySet = <T = unknown>(value: unknown): value is Set<T> => is.set(value)
 is.emptyMap = (value: unknown): value is Map<never, never> => is.map(value) && value.size === 0;
 is.nonEmptyMap = <Key = unknown, Value = unknown>(value: unknown): value is Map<Key, Value> => is.map(value) && value.size > 0;
 
+/**
+`PropertyKey` is any value that can be used as an object key (string, number, or symbol)
+*/
```
gioragutt, line 365:
```suggestion
// `PropertyKey` is any value that can be used as an object key (string, number, or symbol)
```

Sorry for the nitpicking, but comments should either be:

1. oneliners:

```ts
// comment
someCode();
```

or 

2. multiliners:

```ts
/**
 * comment
 */
someCode()
```

Multi-line comments have a certain format to them (each intermediate line starting with `" * "`.

Since there's no real reason to have a multi-line comment, it's better we stick with a single line, IMO.

PopGoesTheWza: My bad, I thought the initial request was for Typescript documentation, hence the `/** ... */` syntax.

If you need multiliner comment but not Typescript documentation, I would suggest the `/* ... */` syntax, with a single leading  `*`.

gioragutt: Did you forget to upload the fix?

gioragutt (changes requested):
The comment about the comment wasn't fixed. I believe this would be the last change

gioragutt (approved):

gioragutt:
@PopGoesTheWza thanks for your contribution bro 🤘🏻

PopGoesTheWza:
> thanks for your contribution bro

On a totally irrelevant and lighthearted matter, I viscerally **hate** the term "bro". I can do with "dude" or "mate" though... ;)

gioragutt:
@PopGoesTheWza I swear I thought about it ASA I posted the comment. Thanks for pointing it out 🙏🏻

Issue #128, opened 2020-12-14
is.iterable returns IterableIterator instead of Iterable


xamgore:
Let's dive into the source code:

https://github.com/sindresorhus/is/blob/d528545e02de3396ea900cd93af478292f0697ee/source/index.ts#L184

Only `[Symbol.iterator]` property is checked, meaning the value is at least `Iterable<T>`. It may be `IterableIterator<T>` if the presence of one more property, `next`, is ensured.

```ts
interface Iterable<T> {
    [Symbol.iterator](): Iterator<T>;
}

interface IterableIterator<T> extends Iterator<T> {
    [Symbol.iterator](): IterableIterator<T>;
}

interface Iterator<T, TReturn = any, TNext = undefined> {
    next(...args: [] | [TNext]): ...;
    return?(value?: TReturn): ...;
    throw?(e?: any): ...;
}
```

Do you agree with it? Would you mind I make a PR?

sindresorhus:
Sounds good. Shouldn't it check all the methods though, not just `next`?

xamgore:
`return` and `throw` are optional, so that's fine.

sindresorhus:
I would describe it as a bug fix.

sindresorhus:
@xamgore Still interested in doing a pull request? No worries if not. Just bumping in case you forgot :)

sdotson:
Taking a crack at this here: https://github.com/sindresorhus/is/pull/149

PR #149, opened 2022-01-06
Fix `is.iterable` and `is.asyncIterable` TypeScript types

sdotson:
Fixes https://github.com/sindresorhus/is/issues/128

Also added `is.iterableIterator`.


sindresorhus:
From the issue:

> And asyncIterable should be fixed too.

sindresorhus:
> Also added is.iterableIterator.

I'm not sure I see the use-case for this one. And if there is one, it should be a separate pull request.

Issue #148, opened 2022-01-03
Proposal: Add support for enum validation


olivierbeaulieu:
Hello!

I've been wondering whether `is` could help me validate that a `string` is a member of a specific enum - I haven't found anything in the library that does that, so I wanted to ask whether that would be a good addition.

Here's what I'm hoping to achieve:

```ts
enum MyEnum {
  SOME_VALUE = 'SOME_VALUE'
}

function myFn(param: string) {
  param; // Type is string
  assert.enum(param, MyEnum);
  param; // Type is MyEnum
}
```

This obviously would not work for const enums - but could be very handy for the rest. Happy to contribute if the proposition makes sense.

sindresorhus:
The example makes no sense to me. Why would you ever accept string if what you actually want is to accept an enum value? Do you maybe have a more realistic example?

In general, I'm lukewarm to adding support for anything regarding TypeScript enums, as [they suck](https://www.aaron-powell.com/posts/2020-05-27-the-dangers-of-typescript-enums/). I love enums in other languages, but the TS one is just bad.

olivierbeaulieu:
I see what you mean - maybe I need to clarify a couple things.

1. We exclusively use non-numeric enums - essentially only strings. Because of the reasons detailed in your first link. But I understand that in designing this library you have to account for the other dangerous cases.
2. The example I gave might be too simplified - what I'm really trying to achieve here is to have an express api handler validate input from params (type will be string), check if the received string is indeed part of the enum of valid values, and cast it as such if the value is valid (and throw if it isn't).

sindresorhus:
Alright. I'm going to accept this.

The check should be `Object.values(Enum).includes(case)`.

I also think `is.enumCase` may be a better name as you're checking the case, not the whole enum itself.

PR #150, opened 2022-01-16
Add `is.enumCase` and `assert.enumCase`

olivierbeaulieu:
Fixes https://github.com/sindresorhus/is/issues/148

Turns out typing this correctly was a bit trickier than I initially expected, since enums are both a value and a type.

Given this:

```ts
enum Direction {
	Asc = 'asc',
	Desc = 'desc'
}
const input: string = 'asc';
assert.enumCase(input, Direction);
```

If `assert.enumCase` was typed to return `asserts value is T` (T being the inferred type of `Direction`), the asserted type of `input` would become `string & typeof Direction`, which is no good.

Typing it as `asserts value is T[keyof T]` instead did the trick, making `input` be typed as `Direction`.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -213,13 +213,13 @@ is.boundFunction(function () {});
 ##### .dataView(value)
 
 ##### .enumCase(value, enum)
-
+**TypeScript-only**
 Returns `true` if `value` is a member of `enum`.
 
 ```ts
-enum EnumDirection {
-	Asc = 'asc',
-	Desc = 'desc'
+enum Direction {
+	Ascending = 'ascending',
+	Descending = 'descending'
 }
 
 is.enumCase('asc', EnumDirection);
--- a/readme.md
+++ b/readme.md
@@ -212,6 +212,23 @@ is.boundFunction(function () {});
 ##### .sharedArrayBuffer(value)
 ##### .dataView(value)
 
+##### .enumCase(value, enum)
+
+Returns `true` if `value` is a member of `enum`.
+
+```ts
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -212,6 +212,23 @@ is.boundFunction(function () {});
...
 
+##### .enumCase(value, enum)
+
+Returns `true` if `value` is a member of `enum`.
+
+```ts
+enum EnumDirection {
+	Asc = 'asc',
+	Desc = 'desc'
+}
```
sindresorhus, line 223:
```suggestion
enum Direction {
	Ascending = 'ascending',
	Descending = 'descending'
}
```

Nitpick

readme.md
```diff
@@ -212,6 +212,23 @@ is.boundFunction(function () {});
 ##### .sharedArrayBuffer(value)
 ##### .dataView(value)
 
+##### .enumCase(value, enum)
+
```
sindresorhus, line 216:
```suggestion
**TypeScript-only**
```

sindresorhus (changes requested):

olivierbeaulieu:
Suggestions, applied, thanks!

papb:
Hello, I think there should have been a test case to verify that `assert.enumCase` really narrows the type, shouldn't it? The added test is runtime-only...

olivierbeaulieu:
I thought about added this in - but I noticed that none of the other test suites did the same, so I left it out. Happy to add this in if the maintainers see the value.

PR #151, opened 2022-01-16
Add type guard for `is.truthy` and `is.falsy`

### LINKED ISSUE #146: Proposal: `is.falsy()` as a type guard
Although there are restrictions on using `is.truthy()` as a type guard function, `is.falsy()` could be a type guard function.

```ts
type Falsy = false | 0 | 0n | '' | null | undefined;

is.falsy = (value: unknown): value is Falsy => !value;
```

zshannon:
Fixes #146 and not sure if adding the same Falsy guard to `is.falsy` helps there because I'm just using `is.truthy` as a `.filter` but happy to add the `Falsy` annotation to `is.falsy` too if helpful

### DIFF
--- a/source/types.ts
+++ b/source/types.ts
@@ -47,3 +47,5 @@ export interface ObservableLike {
 	subscribe(observer: (value: unknown) => void): void;
 	[Symbol.observable](): ObservableLike;
 }
+
+export type Falsy = false | 0 | 0n | '' | null | undefined;
--- a/source/index.ts
+++ b/source/index.ts
@@ -250,7 +250,7 @@ is.urlString = (value: unknown): value is string => {
 // Example: `is.truthy = (value: unknown): value is (not false | not 0 | not '' | not undefined | not null) => Boolean(value);`
 is.truthy = <T>(value: T | Falsy): value is T => Boolean(value);
 // Example: `is.falsy = (value: unknown): value is (not true | 0 | '' | undefined | null) => Boolean(value);`
-is.falsy = (value: unknown) => !value;
+is.falsy = <T>(value: T | Falsy): value is Falsy => !value;
 
 is.nan = (value: unknown) => Number.isNaN(value as number);
 
--- a/source/index.ts
+++ b/source/index.ts
@@ -2,7 +2,7 @@
 /// <reference lib="dom"/>
 /// <reference types="node"/>
 
-import {Class, TypedArray, ObservableLike, Primitive} from './types';
+import {Class, Falsy, TypedArray, ObservableLike, Primitive} from './types';
 
 const typedArrayTypeNames = [
 	'Int8Array',
@@ -247,9 +247,8 @@ is.urlString = (value: unknown): value is string => {
 	}
 };
 
-// TODO: Use the `not` operator with a type guard here when it's available.
 // Example: `is.truthy = (value: unknown): value is (not false | not 0 | not '' | not undefined | not null) => Boolean(value);`
-is.truthy = (value: unknown) => Boolean(value);
+is.truthy = <T>(value: T | Falsy): value is T => Boolean(value);
 // Example: `is.falsy = (value: unknown): value is (not true | 0 | '' | undefined | null) => Boolean(value);`
 is.falsy = (value: unknown) => !value;
 


sindresorhus:
I think it would be better to participate in #146 first. There are unanswered questions there.

source/types.ts
```diff
@@ -47,3 +47,6 @@ export interface ObservableLike {
 	subscribe(observer: (value: unknown) => void): void;
 	[Symbol.observable](): ObservableLike;
 }
+
+
+export type Falsy = false | 0 | '' | null | undefined
```
sindresorhus, line 52: You are missing `0n`.

sindresorhus:
It doesn't look like you ran the tests before submitting.

zshannon:
> It doesn't look like you ran the tests before submitting.

You're right, sorry about that, updated with passing tests. Also added `0n` to `Falsy`.

> Although there are restrictions on using is.truthy() as a type guard function

I don't follow what "restrictions on using is.truthy() as a type guard function" means so not sure how to add to the conversation there. I also don't know "in what cases would [is.falsy as a typeguard] be useful". But `is.truthy` as a typeguard (this PR) is something I use constantly so can easily speak to that if it's unclear :)

sindresorhus:
> I don't follow what "restrictions on using is.truthy() as a type guard function" means 

// @younho9

sindresorhus:
My comment about `is.falsy` clearly applies to `is.truthy` too. The gist of it is "The narrowing is still pretty wide.". How is such a wide narrowing more useful than `unknown`? What can you now use `is.truthy` for that was not possible before the type-guard?

younho9:
I thought `is.truthy` has some restrictions in the following case.

```ts
export type Falsy = false | 0 | 0n | '' | null | undefined;

const isTruthy = <T>(value: T | Falsy): value is T => Boolean(value);
const isFalsy = (value: unknown): value is Falsy => !Boolean(value);

declare const number_: number;

if (isTruthy(number_)) {
    number_ // number
} else {
    number_ // never
}

if (isFalsy(number_)) {
    number_ // 0
} else {
    number_ // number
}
```

`is.falsy` could be useful to narrow falsy value from primitive type, but `is.truthy` is not.

I think using it with a filter could be one of use case.

younho9:
Also, when value has falsy type, `is.truthy` couldn't find truthy value.

```ts
declare const falsyValue: null;

if (isTruthy(falsyValue)) {
    falsyValue // null, expected never
} else {
    falsyValue // never, expected null
}

if (isFalsy(falsyValue)) {
    falsyValue // 0
} else {
    falsyValue // never
}
```

[TypeScript Playground](https://www.typescriptlang.org/play?jsx=0#code/KYDwDg9gTgLgBDAnmYcBiBDANgZ0XAXjgDNsdUAfOABjiuoDs64ByF5hgVyy2c4YAmwYgEsGwAQG4AUNIDGEBjngicAFSicYAC3xEAPGoB8ACgBu2TsABccNc0y5EASlsWsVuKruEjcAEIQEFjAGAzmlsDOMgpKKjiOeIRwER42cPwA1gwQAO4MrnDunt6Jen4AhIHBoeHFUTLSQnJYGFCoscpwXAC2AEbAUAD6tr0DUI0ixCmqGlq6JmODQ87OcADe0nDb3Zz9y3AA9Ie7+1DSAL5wwLiomzun40NHJ+Jmg5eyUzMJZIiLeyeqw2Wx2S2GLxol2utxBD3Bz2Ojw+F1kzVa7TgnXgpCcADVIqNuFhJtMTLNNDp-ri8AS0sD7jsaYg6Z4kVweAAaa7gYByGASbrAd7nK43chwpl-VmodnCwbc0AofmCjlYT7Sb7k35OEzMmUM0HbfWRSFq6Hiu5GkjS01ykWXIA)

zshannon:
I use `isTruthy` as a typeguard with filters everywhere, pattern is like this:
```
type Tipe = { ... }
const maybes: Array<Tipe | undefined | null> = await Promise.allSettled([ ... calls that may fail ... ]).map(({ value }) => value)
maybes.filter(isTruthy).map(X => { ... X is Tipe here ... })
```

I can guess from the defaults in unicorn that @sindresorhus is not a fan of passing eg `isTruthy` directly into `.filter` like this but it's still useful even with the extra `maybes.filter(maybe => isTruthy(maybe))` characters ;)

sindresorhus:
Ah, I get it now. So it's really just about narrowing away `undefined` and `null`.

sindresorhus:
> Also, when value has falsy type, is.truthy couldn't find truthy value.

Do you know why that happens?

sindresorhus:
Seems like this change could be useful even with the mentioned limitations.

sindresorhus:
@zshannon Could you also apply this to `is.falsy`?

PR #152, opened 2022-01-17
Improve `is.integer`, `is.safeInteger`, `is.infinite` TypeScript type

### LINKED ISSUE #147: Proposal: Upgrade type signature for `is.infinite()`, `is.integer()`, `is.safeInteger()`
`is.infinite()`, `is.integer()` and `is.safeInteger()` could provide more strongly-typed number type using `type-fest`.

See https://github.com/sindresorhus/ts-extras/issues/23

younho9:
- It requires `typescript@4.1.2`, since `Integer` type in `type-fest` is using [template literal type](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#template-literal-types)
- Parameter types of `is.integer` and `is.safeInteger` are narrowed to number, because `Integer` type extends number type.
	- It has side effect to `is.evenInteger`, `is.oddInteger`, and also test cases in `is.any` are broken.
	- It could be resolved by making `Integer` type extends unknown.

```ts
export type Integer<Type> = Type extends number ?
	`${Type}` extends `${bigint}`
		? Type
		: never
	: never;
```

resolve #147

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
 		"rxjs": "^6.4.0",
 		"tempy": "^0.4.0",
 		"ts-node": "^8.3.0",
-		"typescript": "~3.8.2",
+		"typescript": "~4.1.2",
 		"xo": "^0.26.1",
 		"zen-observable": "^0.8.8"
 	},
--- a/source/index.ts
+++ b/source/index.ts
@@ -2,7 +2,7 @@
 /// <reference lib="dom"/>
 /// <reference types="node"/>
 
-import {Class, TypedArray, ObservableLike, Primitive} from './types';
+import {Class, TypedArray, ObservableLike, Primitive, Integer} from './types';
 
 const typedArrayTypeNames = [
 	'Int8Array',
@@ -258,8 +258,8 @@ is.nan = (value: unknown) => Number.isNaN(value as number);
 
 is.primitive = (value: unknown): value is Primitive => is.null_(value) || isPrimitiveTypeName(typeof value);
 
-is.integer = (value: unknown): value is number => Number.isInteger(value as number);
-is.safeInteger = (value: unknown): value is number => Number.isSafeInteger(value as number);
+is.integer = <T extends number>(value: T): value is Integer<T> => Number.isInteger(value);
+is.safeInteger = <T extends number>(value: T): value is Integer<T> => Number.isSafeInteger(value);
 
 is.plainObject = <Value = unknown>(value: unknown): value is Record<PropertyKey, Value> => {
 	// From: https://github.com/sindresorhus/is-plain-obj/blob/main/index.js
@@ -279,7 +279,7 @@ export interface ArrayLike<T> {
 	readonly length: number;
 }
 
-const isValidLength = (value: unknown): value is number => is.safeInteger(value) && value >= 0;
+const isValidLength = <T extends number>(value: T): value is Integer<T> => is.safeInteger(value) && value >= 0;
 is.arrayLike = <T = unknown>(value: unknown): value is ArrayLike<T> => !is.nullOrUndefined(value) && !is.function_(value) && isValidLength((value as ArrayLike<T>).length);
 
 is.inRange = (value: number, range: number | number[]): value is number => {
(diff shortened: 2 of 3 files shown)


sindresorhus:
Why are they broken by this change and why would `unknown` fix it?

younho9:
Parameter type of `is.integer` is narrowed to number, so `is.integer` does not extend `Predicate` type.
`is.integer` is used in test cases of `is.any` which require `Predicate` type function, so broken.

Should I just delete these test cases?

https://github.com/sindresorhus/is/blob/main/test/test.ts#L1528
https://github.com/sindresorhus/is/blob/main/test/test.ts#L1553

sindresorhus:
`is.integer()` should still accept `unknown`, like all the other value checks. If it's `unknown`, I guess the type guard can simply be `number`. I don't think we should change the `Integer` type for this though.

younho9:
> `is.integer()` should still accept `unknown`, like all the other value checks. If it's `unknown`, I guess the type guard can simply be `number`. I don't think we should change the `Integer` type for this though.

I agree with your opinion.

sindresorhus:
Why did you close?

younho9:
I thought there

sindresorhus:
Can't you use an overload?

younho9:
Unfortunately, inference from a function will only try the last overload signature.

[TypeScript Playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcCSA7GwDmwoA8AKsqqNhgCYDOcGArgLYBG+AfHALxwAGAJAG8SKAL4845YFVr8BzAJY55WMXAD8cYagBcdYADd8AbgCwAKHMAzehgDGMeRAxx51TNjyEiEkBRp0mVig2AAp9AEMAG3pgXSIASl0I6NRXdCxcfGI2UzNrOwcnFzcMzzComN0bAGsMCAB3DES4ZJjigJZ8OAFzAEgoYBh6KGcAOUD8ADpXd0yocpT43JFzc0pgW0jwgbhbJ2p4Bkit5kjgcc6oXQZLuAAfAOPc83lLEJnS-BCjk7OLoPi8W65jgoMev3OEyg5hEQA)

sindresorhus:
Is this a TypeScript bug/limitation or intentional behavior? If it's a bug/limitation, we could potentially wait for it to be fixed.

sindresorhus:
Could we do something like this? (Not working)

https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcCSA7GwDmwoA8AKsqqNhgCYDOcGArgLYBG+AfHALxwAGAJAG8SKAL4845YFVr8BzAJY55WMXAD8cYagBcdYADd8AbgCwAKHOhIsBKXRZc+APJQAqhgDWGCAHcMxUg5uLQkQCho6JlYodXtsPEItDl0GFmNzcwAzegwAYxh5CAw4eWpMePxiUPDaVOi4AB84HK9fDDYACn0AQwAbemBdIgBKXR7+1FK4xygXd1a-Yg4BcwBIKGAYeihigDko-AA6UvKZrr6B4dMzEQyzSmBc3u6NuFyi6ngGXufmXuB9mkoCkDjEmt9etdzPJMh0Tg4Eh0Id0-gDQcNhnAVmY4LjIj8Uf9AdFzCIgA

younho9:
> Is this a TypeScript bug/limitation or intentional behavior? If it's a bug/limitation, we could potentially wait for it to be fixed.

It seems a design limitation. https://github.com/microsoft/TypeScript/issues/45634#issuecomment-912820467

> Could we do something like this? (Not working)
>
> typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcCSA7GwDmwoA8AKsqqNhgCYDOcGArgLYBG+AfHALxwAGAJAG8SKAL4845YFVr8BzAJY55WMXAD8cYagBcdYADd8AbgCwAKHOhIsBKXRZc+APJQAqhgDWGCAHcMxUg5uLQkQCho6JlYodXtsPEItDl0GFmNzcwAzegwAYxh5CAw4eWpMePxiUPDaVOi4AB84HK9fDDYACn0AQwAbemBdIgBKXR7+1FK4xygXd1a-Yg4BcwBIKGAYeihigDko-AA6UvKZrr6B4dMzEQyzSmBc3u6NuFyi6ngGXufmXuB9mkoCkDjEmt9etdzPJMh0Tg4Eh0Id0-gDQcNhnAVmY4LjIj8Uf9AdFzCIgA

In this example, if value type is string or something, signature could be `function isInteger<string>(value: string): value is number` which return type doesn't match with input type.

sindresorhus:
@younho9 Is there any way we can move this forward or should I close?

younho9:
I couldn't find a way to solve it. I think you'd better close.

PR #159, opened 2022-02-21
Fix `NaN` detection in `.is()`

### LINKED ISSUE #160: [v5] Better handling of `NaN`
https://github.com/sindresorhus/ow/issues/231#issuecomment-1047100612

panva:
`NaN` is not a number, this fixes the is() return for it.

before
```js
is(NaN) // => 'number'
```

after
```js
is(NaN) // => 'NaN'
```

---

Fixes #160

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -50,6 +50,7 @@ const objectTypeNames = [
 	'FormData',
 	'URLSearchParams',
 	'HTMLElement',
+	'NaN',
 	...typedArrayTypeNames
 ] as const;
 
@@ -110,7 +111,7 @@ function is(value: unknown): TypeName {
 		case 'string':
 			return 'string';
 		case 'number':
-			return 'number';
+			return Number.isNaN(value) ? 'NaN' : 'number';
 		case 'boolean':
 			return 'boolean';
 		case 'function':
--- a/test/test.ts
+++ b/test/test.ts
@@ -446,7 +446,7 @@ const types = new Map<string, Test>([
 			NaN,
 			Number.NaN
 		],
-		typename: 'number',
+		typename: 'NaN',
 		typeDescription: AssertionTypeDescription.nan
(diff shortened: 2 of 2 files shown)


panva:
I think this doesn't have to be considered a breaking but will leave that up to you @sindresorhus.

source/index.ts
```diff
@@ -110,7 +111,7 @@ function is(value: unknown): TypeName {
 		case 'string':
 			return 'string';
 		case 'number':
-			return 'number';
+			return Number.isNaN(value) ? 'NaN' : 'number';
```
sindresorhus, line 114: I think it would be clearer to do the `NaN` check on line 107.

panva: That would mean isNan is ran for everything but `null`, this way it is only ran for what we know `NaN` reports as typeof

panva: This optimization can likewise be done for object/null

sindresorhus: You are assuming `Number.isNaN` is not optimized. I feel like this is entering premature optimization territory. In practice, the performance difference is probably none or negligible.

sindresorhus:
The PR looks good, but since I'm planning a major release here in a month or two, I'm gonna keep this PR open until such time, as I don't see the point in risking it breaking for someone.

sindresorhus:
Tests are failing.

panva:
[`Number.isNaN`](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-number.isnan) is not defined to only take an argument of type `number`.

I can either `// @ts-ignore` this or put it back to the `number` case. Or you can go investigate why the types for Number.isNaN are messed up in CI ;)

sindresorhus:
Finally out: https://github.com/sindresorhus/is/releases/tag/v5.0.0

Issue #158, opened 2022-02-20
Proposal: is.nonEmptyStringAndNotWhitespace(value)


ivankatliarchuk:
Currently there is `.emptyStringOrWhitespace(value)`, it would be good to enhance library with `nonEmptyStringAndNotWhitespace(value)`. 

> Returns true if is.nonEmptyString(value) or if it's a string that is all whitespace

Now
```js
.filter((el) => is.nonEmptyString(el) && !is.emptyStringOrWhitespace(el))
```
With negated type
```js
.filter(is.nonEmptyStringAndNotWhitespace)
```

sindresorhus:
Your examples do not do the same thing.

> .filter((el) => is.nonEmptyString(el) && !is.emptyStringOrWhitespace(el))

This check prevents whitespace-only strings

> .filter(s.nonEmptyStringOrWhitespace)

This allows whitespace-only strings. If you read the name.

sindresorhus:
I think what you want is `is.nonEmptyStringAndNotWhitespace`. I'm not sure that's worth it as you could just do:

```js
.filter((el) => !is.emptyStringOrWhitespace(el))
```

ivankatliarchuk:
The proposed solution e.g. `.filter((el) => !is.emptyStringOrWhitespace(el))` does not work.
Example test case https://codesandbox.io/s/npm-playground-forked-u1cypj?file=/src/index.js

1.

```js
const is = require("@sindresorhus/is");

const labels = ["labelA", "", null];
const addLabels = ["labelB", "", undefined, , ,];

function transpose() {
  return [...new Set([...labels, ...addLabels])].filter(
    (el) => is.nonEmptyString(el) && !is.emptyStringOrWhitespace(el)
  );
}

console.log(transpose());
```

`["labelA", "", null, "labelB", undefined]`

2.

```js
const is = require('@sindresorhus/is');

const labels = ['labelA', '', null]
const addLabels =['labelB', '', undefined, , ,]

function transpose() {
  return [...new Set([...labels, ...addLabels])]
    .filter((el) => is.nonEmptyString(el) && !is.emptyStringOrWhitespace(el));
}

console.log(transpose())
```

`["labelA", "labelB"]`

ivankatliarchuk:
Some screenshots as well
![solution-one](https://user-images.githubusercontent.com/5395690/155011663-6027b81e-bb9f-406d-ba35-6dc09884b928.png)
[...]

PR #161, opened 2022-02-22
Add `.nonEmptyStringAndNotWhitespace()`

ivankatliarchuk:
as described here https://github.com/sindresorhus/is/issues/158

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -250,7 +250,7 @@ Returns `true` if the value is a `string` and the `.length` is more than 0.
 Returns `true` if `is.string(value)` and if it's a `string` that is not a whitespace.
 
 ```js
-const values = ['property1', '', null, 'property2', '    ', undefined, , ,]
+const values = ['property1', '', null, 'property2', '    ', undefined]
 
 values.filter(is.nonEmptyStringAndNotWhitespace)
 //=> ['property1', 'property2']
--- a/readme.md
+++ b/readme.md
@@ -247,7 +247,7 @@ Returns `true` if the value is a `string` and the `.length` is more than 0.
 
 ##### .nonEmptyStringAndNotWhitespace(value)
 
-Returns `true` if `is.string(value)` and if it's a `string` that is not a whitespace.
+Returns `true` if `is.string(value)` and if it's a `string` that is non-empty and not a whitespace.
 
 ```js
 const values = ['property1', '', null, 'property2', '    ', undefined]
--- a/readme.md
+++ b/readme.md
@@ -253,7 +253,7 @@ Returns `true` if `is.string(value)` and if it's a `string` that is not a whites
 const values = ['property1', '', null, 'property2', '    ', undefined, , ,]
 
 values.filter(is.nonEmptyStringAndNotWhitespace)
-//=> (2) ['property1', 'property2']
+//=> ['property1', 'property2']
 ```
 
 ##### .emptyStringOrWhitespace(value)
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -245,6 +245,17 @@ Returns `true` if the value is a `string` and the `.length` is 0.
 
 Returns `true` if the value is a `string` and the `.length` is more than 0.
 
+##### .nonEmptyStringAndNotWhitespace(value)
+
+Returns `true` if `is.string(value)` and if it's a `string` that is not a whitespace.
```
sindresorhus, line 250: This description needs work. "not a whitespace" is correct. And it's also missing "non-empty".

readme.md
```diff
@@ -245,6 +245,17 @@ Returns `true` if the value is a `string` and the `.length` is 0.
...
+
+```js
+const values = ['property1', '', null, 'property2', '    ', undefined, , ,]
```
sindresorhus, line 253: The array holes are not relevant.
```suggestion
const values = ['property1', '', null, 'property2', '    ', undefined]
```

readme.md
```diff
@@ -245,6 +245,17 @@ Returns `true` if the value is a `string` and the `.length` is 0.
...
+
+values.filter(is.nonEmptyStringAndNotWhitespace)
+//=> (2) ['property1', 'property2']
```
sindresorhus, line 256:
```suggestion
//=> ['property1', 'property2']
```

test/test.ts
```diff
@@ -1436,6 +1436,22 @@ test('is.emptyStringOrWhitespace', t => {
 	});
 });
 
+test('is.nonEmptyStringAndNotWhitespace', t => {
+	t.false(is.nonEmptyStringAndNotWhitespace(' '));
+	t.false(is.nonEmptyStringAndNotWhitespace(String()));
+	t.true(is.nonEmptyStringAndNotWhitespace('🦄'));
```
sindresorhus, line 1442: You should also test some non-string values

test/test.ts
```diff
@@ -1436,6 +1436,22 @@ test('is.emptyStringOrWhitespace', t => {
...
+
+	t.throws(() => {
+		assert.nonEmptyStringAndNotWhitespace('');
+	});
+	t.throws(() => {
+		assert.nonEmptyStringAndNotWhitespace(String());
+	});
```
sindresorhus, line 1449: This test is not useful.

test/test.ts
```diff
@@ -1436,6 +1436,22 @@ test('is.emptyStringOrWhitespace', t => {
 	});
 });
 
+test('is.nonEmptyStringAndNotWhitespace', t => {
+	t.false(is.nonEmptyStringAndNotWhitespace(' '));
+	t.false(is.nonEmptyStringAndNotWhitespace(String()));
```
sindresorhus, line 1441: This test is not useful.

readme.md
ivankatliarchuk: updated

test/test.ts
ivankatliarchuk: removed

test/test.ts
ivankatliarchuk: Wrong assumption. All code has bugs, I actually expect it to be buggy, and someone can catch an edge case and fix it ;-)

ivankatliarchuk: Added better test coverage

PR #175, opened 2022-11-22
Support assertions custom messages

### LINKED ISSUE #144: Custom error messages
I'm so used to writing context-specific error messages that I wonder how others deal with this because custom messages don't seem to be supported in any of the functions.

```
assert.string(v, `Failed to find a string in this place, instead found value: ${v}`)
```

Maybe I'm missing something about how or when you apply these assertions. Do you use something different when you want to communicate an error back to the user? The standard messages don't tell you much I think unless you're able to at least read the stack trace.

joaompneves:
Add an extra optional param to allow specifying the exception message.

Fixes #144

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,10 +11,8 @@
 		"url": "https://sindresorhus.com"
 	},
 	"type": "module",
-	"exports": {
-		"types": "./dist/index.d.ts",
-		"default": "./dist/index.js"
-	},
+	"exports": "./dist/index.js",
+	"types": "./dist/index.d.ts",
 	"engines": {
 		"node": ">=14.16"
 	},
--- a/source/index.ts
+++ b/source/index.ts
@@ -431,8 +431,12 @@ is.any = (predicate: Predicate | Predicate[], ...values: unknown[]): boolean => 
 
 is.all = (predicate: Predicate, ...values: unknown[]): boolean => predicateOnArray(Array.prototype.every, predicate, values);
 
-const assertType = (condition: boolean, description: string, value: unknown, options: {multipleValues?: boolean} = {}): asserts condition => {
+const assertType = (condition: boolean, description: string, value: unknown, customMessage?: string, options: {multipleValues?: boolean} = {}): asserts condition => {
 	if (!condition) {
+		if (customMessage) {
+			throw new Error(customMessage);
+		}
+
 		const {multipleValues} = options;
 		const valuesMessage = multipleValues
 			? `received values of types ${[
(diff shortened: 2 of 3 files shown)


source/index.ts
```diff
@@ -431,8 +431,12 @@ is.any = (predicate: Predicate | Predicate[], ...values: unknown[]): boolean =>
 
 is.all = (predicate: Predicate, ...values: unknown[]): boolean => predicateOnArray(Array.prototype.every, predicate, values);
 
-const assertType = (condition: boolean, description: string, value: unknown, options: {multipleValues?: boolean} = {}): asserts condition => {
+const assertType = (condition: boolean, description: string, value: unknown, customMessage?: string, options: {multipleValues?: boolean} = {}): asserts condition => {
 	if (!condition) {
+		if (customMessage) {
+			throw new Error(customMessage);
```
sindresorhus, line 437:
```suggestion
			throw new TypeError(customMessage);
```

package.json
```diff
@@ -11,10 +11,8 @@
 		"url": "https://sindresorhus.com"
 	},
 	"type": "module",
-	"exports": {
-		"types": "./dist/index.d.ts",
-		"default": "./dist/index.js"
-	},
+	"exports": "./dist/index.js",
+	"types": "./dist/index.d.ts",
```
sindresorhus, line 15: Don't make unrelated changes.

source/index.ts
```diff
@@ -492,95 +496,95 @@ export const enum AssertionTypeDescription {
 // Type assertions have to be declared with an explicit type.
 type Assert = {
 	// Unknowns.
-	undefined: (value: unknown) => asserts value is undefined;
-	string: (value: unknown) => asserts value is string;
-	number: (value: unknown) => asserts value is number;
-	bigint: (value: unknown) => asserts value is bigint;
+	undefined: (value: unknown, customMessage?: string) => asserts value is undefined;
```
sindresorhus, line 499:
```suggestion
	undefined: (value: unknown, message?: string) => asserts value is undefined;
```

I don't think the "custom" part is needed.

test/test.ts
```diff
@@ -1799,3 +1799,11 @@ test('assert', t => {
 		t.true(is.string(badlyTypedVariable));
 	}
 });
+
+test('assert-custom-message', t => {
```
sindresorhus, line 1803:
```suggestion
test('custom assertion message', t => {
```

test/test.ts
```diff
@@ -1799,3 +1799,11 @@ test('assert', t => {
...
+	const errorMessage = 'Custom error message';
+	const error = t.throws(() => {
+		assert.truthy(false, errorMessage);
+	});
```
sindresorhus, line 1807: Use the `message` option of `t.throws`.

package.json
joaompneves: Unrelated, but current packaging is broken. This puts back to the previous working version, otherwise I couldn't npm install from a git branch.

joaompneves:
@sindresorhus this started as an effort to use this package, but ended up not using it. You can take it from here if you find it relevant.

package.json
sindresorhus: The existing config is totally valid, so I think something else is going on for you.

sindresorhus: Also, if you make unrelated changes, at least comment on why you're doing it. Don't wait for a reviewer to point it out. That just wastes time on both sides.

sindresorhus:
Would have been nice if you had commented that before I spent time reviewing it...

Issue #176, opened 2022-12-07
is.numericString type guarding returns `x is string`  and messes up typescript


Xananax:
Use case:

```js
const parseValue = (/**@type {string} */value) => {
  if (is.numericString(value)) {
    return parseFloat(value);
  }
  if (value.toLowerCase() === "true" || value.toLowerCase() === "false") {
    return value.toLowerCase() === "true";
  }
  //...
}
```

Typescript assumes `value is a string` and then casts `value` to `never`:
![image](https://user-images.githubusercontent.com/836082/206253552-918d54d3-a0e9-48f9-9554-49b7c7f54e2a.png)

I'm not sure if this is solvable, because strictly speaking the type guard is accurate.

Workaround (for anyone who needs it): recast the value manually:

JS:
```js
value = /**@type {string}*/(value)
```
TS:
```ts
value = value as string
```

Notably, this recasting is possible in an early return context like my example, but not in an `if...else` or `switch` context, so it is still a bit of a bother.

The recasting is also surprising and requires a comment for other teammates that may come across it.

sindresorhus:
Maybe we could make the type guard more specific using a technique like this: https://github.com/sindresorhus/type-fest/blob/0b78096186dfe255b888513d60538e17e35828ea/source/internal.d.ts#L108-L111

Xananax:
Oh that sounds pretty good and seems like an easy change. Want me to PR this?

vtgn:
> Use case:
> 
> ```js
> const parseValue = (/**@type {string} */value) => {
>   if (is.numericString(value)) {
>     return parseFloat(value);
>   }
>   if (value.toLowerCase() === "true" || value.toLowerCase() === "false") {
>     return value.toLowerCase() === "true";
>   }
>   //...
> }
> ```
> 
> Typescript assumes `value is a string` and then casts `value` to `never`: ![image](https://user-images.githubusercontent.com/836082/206253552-918d54d3-a0e9-48f9-9554-49b7c7f54e2a.png)
> 
> I'm not sure if this is solvable, because strictly speaking the type guard is accurate.
> 
> Workaround (for anyone who needs it): recast the value manually:
> 
> JS:
> 
> ```js
> value = /**@type {string}*/(value)
> ```
> 
> TS:
> 
> ```ts
> value = value as string
> ```
> 
> Notably, this recasting is possible in an early return context like my example, but not in an `if...else` or `switch` context, so it is still a bit of a bother.
> 
[...]

PR #178, opened 2023-05-30
Discriminate better with `numericString`

Xananax:
This PR allows the `is.numericString` assertion/type guard to not remove `string` from possible remaining values.

Fixes https://github.com/sindresorhus/is/issues/176

I don't believe new tests are necessary.

The assertion
```ts
is `${number}`
```
isn't strictly correct, but I believe it's sufficient.


Xananax:
The tests fail, but they also fail on previous commits with the same crash, so I do not believe my changes are related

PR #180, opened 2023-06-04
fix: revert exports syntax change

### LINKED ISSUE #179: v5.4.0 "Cannot find module '@sindresorhus/is' or its corresponding type declarations"
After upgrading to 5.4.0 I am getting "Cannot find module '@sindresorhus/is' or its corresponding type declarations"

bjornstar:
Resolves #179 

The exports syntax that was used in v5.4.0 only works in node v20+

Reverting it back to the previous syntax fixes the problem, but I tried to test it with node v20 and ran into an error:

```
npx ts-node-esm ./index.ts
TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /Users/bjornstar/test/index.ts
```

This seems to be an issue with ts-node in node v20 - https://github.com/TypeStrong/ts-node/issues/1997
With a WIP PR from isaacs - https://github.com/TypeStrong/ts-node/pull/2009

In the meantime there is a workaround to use the ts-node/esm loader directly:

```
node --no-warnings=ExperimentalWarning --loader ts-node/esm ./index.ts
```

It would be nice if we could unblock existing node v16 / v18 users until `ts-node` is fixed on node v20.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,10 +11,8 @@
 		"url": "https://sindresorhus.com"
 	},
 	"type": "module",
-	"exports": {
-		"types": "./dist/index.d.ts",
-		"default": "./dist/index.js"
-	},
+	"exports": "./dist/index.js",
+	"types": "./dist/index.d.ts",
 	"engines": {
 		"node": ">=14.16"
 	},


sindresorhus:
> The exports syntax that was used in v5.4.0 only works in node v20+

This is incorrect. That syntax has worked since Node.js v12. The problem here is TypeScript.

sindresorhus:
I will revert it for now. But it will come back in the next major version.

PR #181, opened 2023-06-04
Add node v20 to the node version matrix in CI

bjornstar:
Since it's the latest stable version of node, we should include it in our tests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,6 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
+          - 20
           - 18
           - 16
           - 14
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
 	},
 	"scripts": {
 		"build": "del dist && tsc",
-		"test": "tsc --noEmit && xo && ava",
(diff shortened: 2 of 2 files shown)


sindresorhus:
I prefer to wait for `ts-node` to be fixed. Seems like it will happen soon.

bjornstar:
I ran into this again today, can we take this change? It's already been 2 months and there's still no new version of `ts-node` available. Looking at `ava`, they now say [we should use NODE_OPTIONS to specify the loader](https://github.com/avajs/ava/pull/3225)

sindresorhus:
Ok, but I prefer to use the `environmentVariables` AVA config: https://github.com/avajs/ava/blob/main/docs/06-configuration.md

bjornstar:
> Ok, but I prefer to use the `environmentVariables` AVA config: https://github.com/avajs/ava/blob/main/docs/06-configuration.md

Unfortunately, that doesn't work. It looks like [node v20.6.0](https://github.com/nodejs/node/pull/49185) should give us a way to solve our problems, but it's still in progress. It would be nice to have functioning tests for node v20 in the short term to make sure we're not breaking anything. We can remove the ugly config once it's fixed upstream.

PR #182, opened 2023-06-04
Drop node v14 support

bjornstar:
Official support for v14 has [reached end of life at the end of April 2023](https://github.com/nodejs/Release).

We can set node v16 as the minimum version and stop testing on node v14.


sindresorhus:
This will require a major version and I don't see the point in doing that unless there are other useful changes to be done.

PR #187, opened 2023-07-16
Fix type guards for `assert.{truthy,falsy,nan}`

### LINKED ISSUE #186: Some type guards under assert are not working
Several type guards under `assert` are not working.

```ts
import { assert } from '@sindresorhus/is'

function foo(s: string | undefined) {
    assert.truthy(s)
    // Type 'string | undefined' is not assignable to type 'string'.
    // Type 'undefined' is not assignable to type 'string'. ts(2322)
    const bar: string = s
}
```

Nonetheless type guard for `is` is working well.

```ts
import is from '@sindresorhus/is'

function foo(s: string | undefined) {
    if (!is.truthy(s)) {
        throw TypeError(`s is not a string: ${s}`)
    }
    const bar: string = s  // Good!
}
```

This can be verified by looking at [dist/index.d.ts](https://www.npmjs.com/package/@sindresorhus/is?activeTab=code) (line numbers in npm webpage do not align with the code well). For example:

```ts
// dist/index.d.ts (v5.5.0)
declare namespace is {
    var truthy: <T>(value: Falsy | T) => value is T;     // L66
    var falsy: <T>(value: Falsy | T) => value is Falsy;  // L67
}
type Assert = {
    truthy: (value: unknown) => asserts value is unknown;   // L204
    falsy: (value: unknown) => asserts value is unknown;    // L205
}
```

wdzeng:
Fix that type guards for `assert.truthy` and `assert.falsy` are not working since they are declared to be `unknown` type.

Also change type guard for `assert.nan` from `unknown` type to `number` type, as `NaN` is actually of type `number`.

Closes #186

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -294,7 +294,7 @@ is.urlString = (value: unknown): value is string => {
 is.truthy = <T>(value: T | Falsy): value is T => Boolean(value); // eslint-disable-line unicorn/prefer-native-coercion-functions
 
 // Example: `is.falsy = (value: unknown): value is (not true | 0 | '' | undefined | null) => Boolean(value);`
-is.falsy = <T>(value: T | Falsy): value is Falsy => !value;
+is.falsy = (value: unknown): value is Falsy => !value;
 
 is.nan = (value: unknown) => Number.isNaN(value as number);
 
--- a/source/index.ts
+++ b/source/index.ts
@@ -570,9 +570,9 @@ type Assert = {
 	enumCase: <T = unknown>(value: unknown, targetEnum: T) => asserts value is T[keyof T];
 	urlInstance: (value: unknown) => asserts value is URL;
 	urlString: (value: unknown) => asserts value is string;
-	truthy: (value: unknown) => asserts value is unknown;
-	falsy: (value: unknown) => asserts value is unknown;
-	nan: (value: unknown) => asserts value is unknown;
+	truthy: <T>(value: T | Falsy) => asserts value is T;
+	falsy: (value: unknown) => asserts value is Falsy;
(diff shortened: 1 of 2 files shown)


source/index.ts
```diff
@@ -570,9 +570,9 @@ type Assert = {
 	enumCase: <T = unknown>(value: unknown, targetEnum: T) => asserts value is T[keyof T];
 	urlInstance: (value: unknown) => asserts value is URL;
 	urlString: (value: unknown) => asserts value is string;
-	truthy: (value: unknown) => asserts value is unknown;
-	falsy: (value: unknown) => asserts value is unknown;
-	nan: (value: unknown) => asserts value is unknown;
+	truthy: <T>(value: T) => asserts value is Exclude<T, Falsy>;
```
sindresorhus, line 573: Why not use the same type of check as https://github.com/sindresorhus/is/blob/278e0e9696b9862381a6e3cae34f756c67ef735b/source/index.ts#L294 ?

sindresorhus: Can you also add a test to ensure this works correctly?

sindresorhus (changes requested):

source/index.ts
wdzeng: Fixed the types.

May you share some ideas about the unittest? `test/test.ts` are checking runtime behavior, and does not contain tests about whether tsc is able to narrow down types on analyzing.

sindresorhus: https://github.com/sindresorhus/is/blob/278e0e9696b9862381a6e3cae34f756c67ef735b/test/test.ts#L1567-L1575

wdzeng: Done!

PR #190, opened 2023-07-23
Replace enums with string literals

bjornstar:
Replaced the enum `AssertionTypeDescription ` with a type containing all the assertion descriptions.

I'm not sure how much value the string literals have, but the description has a type now instead of `string`.

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -77,6 +77,52 @@ function isPrimitiveTypeName(name: unknown): name is PrimitiveTypeName {
 
 export type TypeName = ObjectTypeName | PrimitiveTypeName;
 
+const assertionTypeDescriptions = [
+	'positive number',
+	'negative number',
+	'Class',
+	'string with a number',
+	'null or undefined',
+	'Iterable',
+	'AsyncIterable',
+	'native Promise',
+	'EnumCase',
+	'string with a URL',
+	'truthy',
+	'falsy',
+	'primitive',
+	'integer',
+	'plain object',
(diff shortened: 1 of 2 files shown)


sindresorhus:
What problem does it solve? This would be a breaking change as the existing type is exported.

bjornstar:
The arguments are the same as #113, I believe the only reason the `AssertionTypeDescription`s are exported is for the tests. Doing a search for `AssertionTypeDescription` across all of github does not turn up any usage.

I understand not wanting to make a breaking change for what may seem like cosmetic reasons. Changing a export into a type export reduces the footprint of the interface while improving the developer experience.

I have some spare time at the moment and my intention is to make more substantial improvements to the library, specifically doing named exports and improving the type assertion descriptions for `is.any`. I feel like it might be a bit much for a single PR and I wanted to get the enums out of the way. Do you have a preference for how I should proceed?

sindresorhus:
Thanks for elaborating. I agree, this is better and it's unlikely anyone is depending on the enum.

sindresorhus:
> I have some spare time at the moment and my intention is to make more substantial improvements to the library, specifically doing named exports and improving the type assertion descriptions for is.any. I feel like it might be a bit much for a single PR and I wanted to get the enums out of the way. Do you have a preference for how I should proceed?

Improvements in that area are very welcome. They should separate pull requests. The named export pull request will cause large changes, so I would do that one first, and wait with any other changes until that is merged, to avoid conflicts.

PR #191, opened 2023-08-02
Implement named exports

### LINKED ISSUE #137: ESM conversion plan: what exports?
What will this module export as a ESM? Will it still be a single default object like it is now or will it export `isUndefined` and the like? 

It could potentially use both:

``` js
export function isUndefined(){}
const is = {
  undefined: isUndefined
}
export default is;
```

bjornstar:
Resolves #137 

I added named exports for every method, both is and assert.

While implementing it, I exported validLength and whitespaceString as they were both used internally but seemed useful.

I also couldn't stop myself from improving the type for isInRange to be `[number, number]` instead of `number[]`. I can revert this change if you'd like.

I also alphabetized the method and function declarations for easier detection of missing entries.

One nice side effect of this process is that we can remove the method names that were renamed because they were reserved words in typescript. I left them in for backwards compatibility but marked them as deprecated.

I moved the ArrayLike, NodeStream, and Predicate type definitions into the types declaration, they seem to fit in better there. I can move them back if you prefer.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -54,11 +54,11 @@ assert.string(foo);
 // `foo` is now typed as a `string`.
 ```
 
-## Named Exports
+### Named exports
 
 Named exports allow tooling to perform tree-shaking, potentially reducing bundle size by including only code from the methods that are used.
 
-Every method listed above is available as a named export. Each method is prefixed by either `is` or `assert` depending on usage.
+Every method listed below is available as a named export. Each method is prefixed by either `is` or `assert` depending on usage.
 
 For example:
 
--- a/readme.md
+++ b/readme.md
@@ -560,6 +560,8 @@ is.all(is.string, '🦄', [], 'unicorns');
 
 Returns `true` if the value is a safe integer that is greater than or equal to zero.
 
+This can be useful to confirm that a value is a valid count of something, ie. 0 or more.
+
 ##### .whitespaceString(value)
 
 Returns `true` if the value is a string with only whitespace characters.
--- a/readme.md
+++ b/readme.md
@@ -72,11 +72,17 @@ Example:
 - `'Function'`
 - `'Object'`
 
+This method is also exported as `detect`, you can import it like this:
+
+```js
+import {detect} from '@sindresorhus/is';
+```
+
 Note: It will throw an error if you try to feed it object-wrapped primitives, as that's a bad practice. For example `new String('foo')`.
 
 ### is.{method}
 
-All the below methods accept a value and returns a boolean for whether the value is of the desired type.
+All the below methods accept a value and return a boolean for whether the value is of the desired type.
 
 #### Primitives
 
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -554,6 +548,14 @@ is.all(is.string, '🦄', [], 'unicorns');
 //=> false
 ```
 
+##### .validLength(value)
+
+Returns `true` if the value is a safe integer that is greater than or equal to zero
+
+##### .whitespaceString(value)
+
+Returns `true` if the value is a string with only whitespace characters
```
sindresorhus, line 557:
```suggestion
Returns `true` if the value is a string with only whitespace characters.
```

sindresorhus (changes requested):

sindresorhus:
One downside of having the default export `is` be a function is that we cannot re-export all the named exports as a default export, making them tree-shakable even when using the default export. I wonder if it's worth moving the function to something like `is.detect()`. Thoughts?

I have done something similar here: https://github.com/sindresorhus/query-string/blob/main/index.js

bjornstar:
> One downside of having the default export `is` be a function is that we cannot re-export all the named exports as a default export, making them tree-shakable even when using the default export. I wonder if it's worth moving the function to something like `is.detect()`. Thoughts?

It makes sense to provide access to the original method through a named export. I had been struggling to find a name for it, `detect` works for me.

I think it would be nice to allow users to continue using the default export, at least for a little bit to give them time to migrate completely to named exports.

bjornstar:
I have updated the readme and exported the default method as `detect`.

readme.md
```diff
@@ -554,6 +562,28 @@ is.all(is.string, '🦄', [], 'unicorns');
...
+
+Named exports allow tooling to perform tree-shaking, potentially reducing bundle size by including only code from the methods that are used.
+
+Every method listed above is available as a named export. Each method is prefixed by either `is` or `assert` depending on usage.
+
+For example:
+
+```js
+import {assertNull, isUndefined} from '@sindresorhus/is';
+```
```
sindresorhus, line 585: I think this should go at the end of the `Usage` section

bjornstar: Good idea, I moved it up underneath the Usage section.

sindresorhus:
Nice work 👍

Do you plan more changes or should I release this?

bjornstar:
> Nice work 👍
> 
> Do you plan more changes or should I release this?

I'd like to do the `is.any` / `is.all` assertion descriptions.

It would also be nice to include [dropping node v14](https://github.com/sindresorhus/is/pull/182) since this is a breaking change, can I reopen #182?

sindresorhus:
> I'd like to do the is.any / is.all assertion descriptions.

👍 

> It would also be nice to include https://github.com/sindresorhus/is/pull/182 since this is a breaking change, can I reopen https://github.com/sindresorhus/is/pull/182?

👍

PR #193, opened 2023-08-07
Give better assertion messages for assert.all & assert.any

### LINKED ISSUE #107: Enhance assert error messages
`is.assert.any`:

https://runkit.com/szmarczak/5e53ec822c7e080014a4edf9

```
TypeError: Expected value which is `predicate returns truthy for any value`, received value of type `Array`.
```

IMO it should return:
```
Expected value of type `boolean` or `undefined`, got `null`
```

and if the check function is not an `is` function, simply:
```
Expected value of a custom type, got `null`
```

bjornstar:
Fixes #107 

I improved the output slightly for the type assertion messages and added tests for the improvements.


source/index.ts
```diff
@@ -794,10 +794,33 @@ function typeErrorMessage(description: AssertionTypeDescription, value: unknown)
...
+		case 1: {
+			return values.join('');
+		}
+
+		case 2: {
+			return values.join(` ${word} `);
+		}
+
+		default: {
+			// TODO: Replace with .at after node v14 support is dropped
```
sindresorhus, line 814: This can be done now.

source/index.ts
```diff
@@ -794,10 +794,33 @@ function typeErrorMessage(description: AssertionTypeDescription, value: unknown)
...
 
-function typeErrorMessageMultipleValue(description: AssertionTypeDescription, values: unknown[]): string {
+function unique<T>(values: T[]): T[] {
 	// eslint-disable-next-line unicorn/prefer-spread
-	const valueTypes = Array.from(new Set(values.map(singleValue => `\`${is(singleValue)}\``))).join(', ');
-	return `Expected value which is \`${description}\`, received values of types ${valueTypes}.`;
+	return Array.from(new Set(values));
+}
+
+function joinWithWord(values: string[], word: 'and' | 'or'): string {
```
sindresorhus, line 802: Maybe use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat

bjornstar: Good idea, I switched to using `Intl.ListFormat` but ran into an issue where the type definitions did not exist for it.

I updated `tsconfig.json` to target ES2021.

PR #196, opened 2023-08-10
Rename domElement to htmlElement

bjornstar:
It seems like `isHtmlElement` is a better description for a method that determines if something is of type `HTMLElement`.

I didn't export a copy of `isHtmlElement` as `isDomElement`, I can add it if you'd prefer.


sindresorhus:
> I didn't export a copy of isHtmlElement as isDomElement, I can add it if you'd prefer.

Nah, not needed.

PR #197, opened 2023-08-22
Add React Native compatibility

DucAnhPhi:
- add missing `main` entry in `package.json`, leading to not found index file (this happened in RN project)
- remove `Intl.ListFormat` as it is not yet supported by Hermes (JS engine for RN), see: https://github.com/facebook/hermes/blob/main/doc/Features.md#planned

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
 		"url": "https://sindresorhus.com"
 	},
 	"type": "module",
+        "main": "./dist/index.js",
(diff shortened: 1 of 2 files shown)


sindresorhus:
No thanks :)

This package targets Node.js 16 and modern browsers, where it works perfectly. I'm not interested in changing anything to make it work with React Native.

PR #198, opened 2023-08-23
Add main entry to package.json

DucAnhPhi:
- without the main entry, the index file cannot always be found implicitly

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
 		"url": "https://sindresorhus.com"
 	},
 	"type": "module",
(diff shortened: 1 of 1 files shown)


sindresorhus:
> without the main entry, the index file cannot always be found implicitly

`exports` is the modern way to define main entry. This package has the correct config. Any problems are caused by bundlers and other tools.

https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

DucAnhPhi:
thanks, good to know

PR #203, opened 2024-02-08
Add `.validDate()`

marlun78:
This PR adds a new `.validDate()` function. It returns `true` if the passed value is a valid date. 

`Invalid Date`s occur when an invalid value is passed to the `Date` constructor. This can be an invalid date string, a non-numeric value or a number that is outside of the expected range for a date value. `Invalid Date`s are tecnically `Date`s, but not very useful. 

For more details on invalid `Invalid Date`s, see [_The epoch, timestamps, and invalid date_ on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_epoch_timestamps_and_invalid_date).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -578,13 +578,29 @@ is.all(is.string, '🦄', [], 'unicorns');
 
 Returns `true` if the value is a valid date.
 
-`Invalid Date`s occur when an invalid value is passed to the `Date` constructor. This can be an invalid date string, a non-numeric value or a number that is outside of the expected range for a date value.
+All [`Date`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date) objects have an internal timestamp value which is the number of milliseconds since the [Unix epoch](https://developer.mozilla.org/en-US/docs/Glossary/Unix_time). When a new `Date` is constructed with bad inputs, no error is thrown. Instead, a new `Date` object is returned. But the internal timestamp value is set to `NaN`, which is an _'Invalid Date'_. Bad inputs can be an non-parsable date string, a non-numeric value or a number that is outside of the expected range for a date value.
 
 ```js
-is.validDate(new Date());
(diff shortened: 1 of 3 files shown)


marlun78:
Any ideas why the timeout?

test/test.ts
```diff
@@ -2101,6 +2101,17 @@ test('is.urlSearchParams', t => {
...
 
+test('is.validDate', t => {
+	t.true(is.validDate(new Date()));
+	t.false(is.validDate(new Date('x')));
+	t.notThrows(() => {
+		assert.validDate(new Date());
+	});
+	t.throws(() => {
+		assert.validDate(new Date('x'));
+	});
```
sindresorhus, line 2112: Needs a lot more assertions.

readme.md
```diff
@@ -574,6 +574,20 @@ is.all(is.string, '🦄', [], 'unicorns');
 //=> false
 ```
 
+##### .validDate(value)
+
+Returns `true` if the value is a valid date.
+
+`Invalid Date`s occur when an invalid value is passed to the `Date` constructor. This can be an invalid date string, a non-numeric value or a number that is outside of the expected range for a date value.
```
sindresorhus, line 581: Should more clearly describe that a `Date` instance can be in an invalid state, meaning it's possible to create an invalid `Date` without the constructor throwing. I don't think that's immediately clear to all users.

sindresorhus (changes requested):

test/test.ts
marlun78: Thanks for the feedback @sindresorhus!

Please help me understand what you are after. Do you feel like we needs more assertions to serve as examples/documentation? I mean, a date can only ever be in two states, either valid (represented by a number) or invalid (represented by `NaN`), so we can't really test it in more than two ways.

sindresorhus:
Thanks :)

PR #204, opened 2024-04-19
Support custom assertion messages

marlun78:
Adds support for custom assertion messages. Unfortunately, the variadic functions `assertAll` & `assertAny` cannot accept a message without a breaking change. 

Closes https://github.com/sindresorhus/is/issues/144.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -45,6 +45,15 @@ assert.string(2);
 //=> Error: Expected value which is `string`, received value of type `number`.
 ```
 
+Most assertions also support an optional custom error message.
+
+```js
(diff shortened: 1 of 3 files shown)


sindresorhus:
Can you add an example using the message parameter here: https://github.com/sindresorhus/is/blob/664b9077e1666531e359118aa5ae1cea5983a2dd/readme.md#L46 Just to make it obvious to non-TS users that it's supported.

sindresorhus:
Thanks :)

0x80:
Thanks a for adding this!

PR #215, opened 2025-06-05
Structure fixtures so they can be easily tested for exclusivity 

bjornstar:
Without extensive exclude declarations as specified in the TODO message:

```
```

I identified each unique entry in the fixtures and, if they were reused across multiple types, hoisted them to the top and spread them into the fixtures. This allows us to filter out fixture values that have already been tested as being of the type under test. It becomes more declarative in the fixture definitions, rather than in the call site of the test.

I split the primitive type fixtures from the object type fixtures so that we can exclude all the entries that are not primitives without needing redundant properties in the fixture data.

Another structural change to the fixtures was changing the fixtures from a Map to a readonly object so that we can use the keys to access the `is` methods that we are testing. This removed 2 redundant lines from each fixture.

I also iterate over the method names defined in the fixtures to ensure that everything listed is tested. Previously every fixture was used manually and needed to be added to a test after it was created.

If a method name has additional tests beyond the exclusive fixture based test, I labelled it as supplemental.

There was one tricky bit in there with `Buffer` and `Uint8Array` being the same thing so I added an exclusion there.

⚠️ There was one small functional change that I added regarding the Promise type. I think we should prefer the `Promise` type over `Object` when it has the promise API. It seems strange that we would expect the typeDescription to be `Promise`, but the typeName would be `Object`. If you think we should leave it as is, I can probably create an exclusion for it.

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -209,8 +209,12 @@ function detect(value: unknown): TypeName {
 	}
 
 	const tagType = getObjectType(value);
-	if (tagType) {
+	if (tagType && tagType !== 'Object') {
 		return tagType;
+	}
+
+	if (hasPromiseApi(value)) {
+		return 'Promise';
 	}
 
 	if (value instanceof String || value instanceof Boolean || value instanceof Number) {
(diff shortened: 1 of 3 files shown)


source/index.ts
```diff
@@ -1120,7 +1124,7 @@ const methodTypeMap = {
 	isWhitespaceString: 'whitespace string',
 } as const;
 
-function keysOf<T extends Record<PropertyKey, unknown>>(value: T): Array<keyof T> {
+export function keysOf<T extends Record<PropertyKey, unknown>>(value: T): Array<keyof T> {
```
sindresorhus, line 1127: This should not be exposed to users. Would be better to move it to a `utilities` file and import it into both here and tests.

bjornstar: Done!

sindresorhus:
Thanks :)

